Issue #518, opened 2026-07-26
Epic: Strict & exact regex test coverage for all structural signatures, per language


squid-protocol:
## Why

`gitgalaxy/standards/language_standards.py` defines **2,182 compiled regex rules** across **58 languages** — the Structural Signature Analysis Engine that is the core of how GitGalaxy parses code without an AST. This epic originally started from the premise that **6 languages** (C, C++, C#, COBOL, Rust, TypeScript) already had dedicated correctness/ReDoS testing (`tests/core_engine/test_language_standards_strict.py`) and so were out of scope. Every other language's rules were exercised only by a blanket "does it compile" fuzzer (`test_global_regex_syntax_integrity`) — which proves the regex is syntactically valid Python, not that it matches (or correctly *excludes*) what its own documented semantics say it should.

**Update (2026-07-29): that premise was checked and doesn't hold up.** Auditing what those 6 languages actually have in the test file: each one has only a **handful of isolated regression tests** (`test_c_knr_ambiguity_trap`, `test_csharp_iron_wall_redos`, `test_cpp_macro_multiline_spiral`, `test_c_pointer_ambiguity_overlap`, `test_cobol_ghost_satellite_prevention`, `test_thermodynamic_operator_collisions`), each one hard-coded to a single specific bug someone already found — not the full per-signature template (positive/negative case + ambiguity sweep + ReDoS scaling sweep for every non-`None` signature) every other language in this epic has now received. Coverage was roughly 2-12% of each language's actual non-`None` signature count (e.g. COBOL: 1 of 46 signatures touched; TypeScript: 1 of 58). **Filed as 6 new sub-issues** (#773-778).

## Update (2026-07-30): epic complete — all 6 newly-filed languages closed

Every one of the 6 languages found to have thin coverage now has the full per-signature template, in the same style as the other 48:

- **typescript** (#778 → #779): 58 signatures, **8 real bugs**.
- **rust** (#777 → #781): 51 signatures, **12 real bugs** — the largest single-language count in the epic at the time, driven almost entirely by one root cause repeated across 10 rules (`!`-suffixed macro invocations like `println!`/`panic!`/`asm!` sharing a trailing `\b` with word-ending siblings).
- **cobol** (#776 → #783): 46 signatures, **4 real bugs** + a missing `import` schema key found via a full key-list diff, not a pattern defect.
[...]

PR #1026, opened 2026-08-04
Groovy: fix return-type/control-flow/nested-paren bugs in args regex

squid-protocol:
Recovered from an abandoned background-agent worktree -- never committed at the time. Three real bugs in the "args" pattern:

- Lowercase primitive return types (void foo(), int add()) never matched since the return-type stepper required an uppercase first char. Added a primitive-keyword alternative alongside the existing capitalized-type branch, and allowed `,` so multi-param generic return types (Map<String, Integer> foo()) don't break after the first type param.
- The flat [^)]* param-list matcher truncated at the first inner `)`, breaking on one-level-nested calls in default values (int y = Math.max(3, 4)). Swapped for a bounded one-level-nesting form.
- No exclusion for control-flow keywords, so if (x) {, while (x) {, switch (x) {, for (i in ...) {, catch (Exception e) {, and synchronized(lock) { were all misread as method parameter blocks. Added the same negative-lookahead style func_start already uses.

Not yet verified against the golden-master corpus or opened as a PR -- needs the usual crucible_check.py pass before merging.

### Visual Observatory Testing
- [ ] I tested the output JSON on **GitGalaxy.io** OR the **Airgap Observatory**.
- [ ] Flexbox constraints, HUD elements, and 3D rendering remain intact.

### The Differential Scan Acknowledgement
- [ ] I understand that my PR will be subjected to a Full Differential Scan.
- [ ] I have provided the link to the specific repository this PR addresses so it can be tested alongside the 80-repo calibrated baseline.
- [ ] I believe these changes will measurably improve the engine's Accuracy, Speed, Utility, or Ethos without causing regressions.

### DIFF
--- a/gitgalaxy/standards/language_standards.py
+++ b/gitgalaxy/standards/language_standards.py
@@ -10687,8 +10687,30 @@ LANGUAGE_DEFINITIONS = {
             # (`[a-zA-Z_$][\w_$]*` with no \b anchor) got retried at every
             # position in a long `->`-less line -- O(n^2) (same bug found and
             # fixed in javascript's args). Bounded to {0,100}.
+            # BUG FIX: the return-type stepper required an uppercase first
+            # char, so lowercase primitive/void return types (`void foo()`,
+            # `int add()`) never matched at all. It also excluded `,`, so
+            # multi-param generic return types (`Map<String, Integer> foo()`)
+            # broke after the first type param. Added a primitive-keyword
+            # alternative and `,` to the class (same technique already used
+            # by C#'s func_start for this exact bug class).
+            # NESTED-DELIMITER FIX: the flat `[^)]*` param-list matcher broke
+            # on one-level-nested calls in default values (`int y =
+            # Math.max(3, 4)`), truncating at the inner `)`. Swapped for a
+            # bounded one-level-nesting form.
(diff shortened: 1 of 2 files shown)


squid-protocol:
Verified against current `main` — all three bugs described here are already fixed. They were fixed and merged in PR #712 (2026-07-28, closing #584 as part of epic #518), which predates this issue's creation (2026-08-04):

- Lowercase primitive return types (`void foo()`, `int add()`): fixed via the primitive-keyword alternative in the return-type stepper (`gitgalaxy/standards/language_standards.py` args regex, ~L12204), plus `,` added to the identifier class for multi-param generics.
- One-level-nested parens in default values (`Math.max(3, 4)`): fixed via the bounded `(?:[^()]|\([^()]*\))*` nesting form.
- Control-flow keywords misread as param lists (`if`, `while`, `switch`, `for`, `catch`, `synchronized`): fixed via the negative lookahead `(?!(?:if|for|while|switch|catch|synchronized)\b)`.

Confirmed with a local run: all 144 tests in `tests/extraction/languages/test_groovy_strict.py` + `test_groovy.py` pass, including dedicated regression tests for each of these three bug classes (e.g. `test_groovy_strict.py:259`, `:285`, `:345`).

This issue was recovered from an abandoned background-agent worktree that apparently never synced with the epic #518 work landing on main. Closing as already fixed / duplicate of #584.

Issue #1041, opened 2026-08-04
bug(core): Nested functions are silently dropped from extraction


squid-protocol:
## Problem

Two of the four function-slicing strategies in `gitgalaxy/core/detector.py` contain identical guards that cause nested/inner functions to be silently dropped:

**Mode B (brace languages)** — `_slice_by_braces` at line 1564:
```python
start_idx = match.start()
if start_idx < last_end_idx:
    continue
```

**Mode C (indentation languages)** — `_slice_by_indentation` at line 1693:
```python
start_idx = match.start()
if start_idx < last_end_idx:
    continue
```

`last_end_idx` tracks the end index of the previously accepted function match (set at line 1640 for braces, line 1733 for indentation).

## Bug

A nested/inner function declaration (e.g. a Python function defined inside another function, or a JavaScript function expression assigned inside another function's body) necessarily starts *before* its enclosing function's end index. This guard causes it to **always be skipped** — it never becomes its own `FunctionNode`.

**Consequence**: 
- The source text of the nested function is still included as part of the parent function's `block` (so its complexity signals bleed into the parent's aggregate `hit_vector`/`branch_count`/etc.)
- But it is never counted, named, or reported as a distinct function
- `function_count` is undercounted for any file using nested functions/closures
- Per-function complexity for the true innermost logic is invisible (merged into the outer function's numbers instead)
- Any future per-method analysis (e.g. LCOM within a class whose methods contain nested helper functions) would misattribute the nested function's behavior to the outer method

## Example

```python
def outer():
    def inner():      # SILENTLY DROPPED — never becomes its own FunctionNode
        return 42
    return inner()
```

Detector sees:
- 1 function total (outer)
- Inner's branch/logic complexity merged into outer's hit_vector

Should see:
- 2 functions (outer, inner)
- Inner's metrics separate from outer

## Fix Location

- Mode B (braces): `gitgalaxy/core/detector.py` line 1564 guard + line 1640 `last_end_idx` update
- Mode C (indentation): `gitgalaxy/core/detector.py` line 1693 guard + line 1733 `last_end_idx` update
- Mode A (labels/assembly) at line 1775+ does NOT have this guard and is not affected the same way

## Fix Direction

The fix needs a proper nesting-depth-aware or stack-based scope tracker rather than the current flat "skip if inside previous match's span" approach. The guard exists specifically to prevent double-counting, but has the side effect of dropping real nested definitions entirely instead of counting them as their own (nested) function.

[...]

PR #1045, opened 2026-08-05
Fix nested functions being silently dropped from extraction (#1041)

squid-protocol:
## Summary

- Mode B (`_slice_by_braces`) and Mode C (`_slice_by_indentation`) in `gitgalaxy/core/detector.py` both had an identical guard -- `if start_idx < last_end_idx: continue` -- that skipped any `func_start` match whose start fell before the previously accepted match's end. A nested/inner function declaration necessarily starts before its enclosing function's end, so this guard silently dropped every nested function instead of ever making it its own `FunctionNode` (undercounting `function_count`, and merging the nested function's complexity into its parent's aggregate metrics instead of reporting it separately).
- The guard was unnecessary: each match already resolves its own end independently from its OWN start position (brace-depth tracking via `_find_balanced_end` for Mode B, dedent tracking for Mode C) -- so once a nested match is allowed to be processed at all, it already gets a correctly bounded, correctly nested scope with no further changes needed. Removing the guard (and the now-dead `last_end_idx` bookkeeping) is the whole fix.
- Mode A (labels/assembly) never had this guard and is unaffected, per the issue.

## Verification

- Repro'd the bug pre-fix (Python `def`-in-`def` and a brace-style nested function both collapsed to 1 extracted function instead of 2), confirmed fixed post-fix, including a sibling-nested-functions case (two inner defs at the same nesting depth under one outer function).
- Added regression tests: `test_detector_nested_function_is_counted_as_own_node_braces` and `test_detector_nested_function_is_counted_as_own_node_indentation` in `tests/core_engine/test_detector.py`.
- Full local suite: `5435 passed, 2 skipped, 1 deselected, 11 xfailed, 3 xpassed`.
- `tests/tools/audit_check.py --regenerate`: ruff baseline updated for 2 pure line-shift findings from the removed dead code (no new findings); mypy and dead-key audits clean.

## Note for reviewers

This is a parsing-logic change to `detector.py`, so per the repo's Differential Scan protocol it should be checked against `tests/golden_master_audit.json` / `tests/golden_master_zero_dep_audit.json` by CI's `crucible-audit` check against the ~80-repo corpus. I don't have `LANGUAGE_CRUCIBLE_PATH` set up locally, so I did not run `crucible_check.py` -- deferring to CI here per the repo's cost-conscious-verification convention. Function counts are expected to increase for any corpus file with nested/inner functions (that's the intended fix), so if `crucible-audit` flags a diff, it should be reviewed and the golden master re-blessed via `update_golden_master.py` rather than treated as a regression by default -- happy to do that once CI reports the actual diff.

## Test plan

- [x] `pytest tests/core_engine/test_detector.py -q` -- 82 passed
- [x] `pytest tests/ -q` -- full suite green
- [x] `tests/tools/audit_check.py` -- clean after baseline regen
- [ ] CI `crucible-audit` -- pending, may require golden master re-bless (see note above)


squid-protocol:
Re-blessed the golden master fixtures against the language-crucible corpus (`crucible_check.py --update`, both full-precision and zero-dependency modes) -- see the follow-up commit for the full rationale. \`crucible_check.py\` now passes clean in both modes locally; pushing to let CI's \`crucible-audit\` confirm.

Issue #83, opened 2026-06-25
Build SARIF Exporter for Native GitHub/GitLab Integration


squid-protocol:
**Epic:** #75
**Target:** `gitgalaxy/recorders/sarif_recorder.py` & `galaxyscope.py`

## Context & Strategic Value
To be adopted by enterprise engineering teams, GitGalaxy cannot force users to learn a new dashboard. We must bring our findings directly to the developer's pull request. By exporting our Structural Signatures and Risk Vectors into the OASIS-compliant SARIF format, GitHub Advanced Security (GHAS), GitLab, and Azure DevOps will automatically ingest our telemetry and annotate code in real-time.

## Implementation Tasks

- [ ] Create `gitgalaxy/recorders/sarif_recorder.py`.
- [ ] Initialize the SARIF JSON structure with the required `version: "2.1.0"` and `runs` array.
- [ ] Map GitGalaxy telemetry into the `tool.driver.rules` array. 
    - *Example: A `high_risk_execution` spike becomes a formalized rule object.*
- [ ] Map spatial `hit_vector` coordinates into SARIF `physicalLocation` blocks.
    - *Note: You must translate regex byte offsets or matched indices into actual line numbers (URI and `startLine`/`endLine`).*
- [ ] **[NATIVE INTEGRATION]** Update `galaxyscope.py` to import `SarifRecorder`. Append it to the final export block alongside JSON and SQLite to generate `gitgalaxy_results.sarif` by default.

## Validation
Run a local scan generating the `.sarif` file. Upload it to a test GitHub repository using the `github/codeql-action/upload-sarif` action. Verify that GitGalaxy risk findings appear natively under the "Security" tab and inline on PR diffs.

PR #1185, opened 2026-08-09
fix(packaging): remove dead zero-trust-sbom console-script entry point (#1179)

### LINKED ISSUE #1179: Fix dead zero-trust-sbom console-script entry point (module doesn't exist)
## 🛑 Problem Statement

`pyproject.toml`'s `[project.scripts]` registers:

```
zero-trust-sbom = "gitgalaxy.tools.compliance.sbom_generator:main"
```

`gitgalaxy/tools/compliance/` doesn't exist anywhere in the repo — there is no `compliance/`
directory at all, let alone a `sbom_generator.py` inside it. Anyone who installs GitGalaxy via
PyPI and runs `zero-trust-sbom` gets an immediate `ModuleNotFoundError` (or the entry point fails
to resolve at all). Found while evaluating the `gitgalaxy/security/` vs. `gitgalaxy/tools/`
organization (unrelated to that work, just tripped over it in the same pass).

Real SBOM generation exists and works today, just under a different name and invocation path:
`gitgalaxy/recorders/sbom_recorder.py`'s `SbomRecorder` class, invoked as part of the main
`galaxyscope`/`blast` entry point via the `--sbom-only` exclusive-mode flag (see
`galaxyscope.py`'s `SBOM_ONLY` config handling, `--sbom-only` argparse flag, and
`generate_report(...)` call around line ~1319). `sbom_recorder.py` itself has no `main()`/
standalone entry point — it was never built as an independently invocable module.

## 🎯 Objective

Pick one of two real fixes (this issue is deciding between them, not prescribing one):

1. **Remove the dead entry.** `galaxyscope --sbom-only` already does exactly what
   `zero-trust-sbom` was presumably meant to do — the standalone command may just be redundant.
2. **Make it real.** Add a thin `main()` wrapper (could live at
   `gitgalaxy/tools/compliance/sbom_generator.py`, actually creating that module this time, or
   more simply as a small function in an existing file) that just calls `galaxyscope.main()`
   with `--sbom-only` forced — the same pattern `blast` already uses as a pure alias for
   `galaxyscope:main`. This gives users a dedicated, discoverable `zero-trust-sbom` command
   without typing the full `--sbom-only` flag.

Lean toward option 2 if `zero-trust-sbom` is referenced anywhere user-facing (README, docs,
marketing copy) as an existing/expected command — check before deciding, since removing a
documented command is a bigger compatibility change than fixing a broken one.

## 🛠️ Proposed Implementation

- Grep the repo (README.md, docs/, GitHub Actions composite action config) for existing
  `zero-trust-sbom` references to see whether anything already documents/depends on this name.
- Implement whichever of the two options above the grep points to.
- If keeping it: verify `pip install -e . && zero-trust-sbom .` actually produces a valid SBOM
  output, not just that the entry point resolves.

## ✅ Acceptance Criteria

- [ ] `zero-trust-sbom` either produces a working SBOM when run after a fresh install, or is
      removed from `pyproject.toml` with a clear reason recorded
- [ ] No other `pyproject.toml` console-script entries point at nonexistent modules (quick sweep
      of the full `[project.scripts]` table while in here, since this one was found by accident —
      worth checking the rest aren't also stale)

ringoliRob:
…t (#1179)

`zero-trust-sbom` pointed at `gitgalaxy.tools.compliance.sbom_generator:main`, a module that has not existed since 638c8a2 (#83). Anyone installing from PyPI and running the command got a failure to resolve the entry point.

That commit deliberately migrated the CycloneDX generator out of the standalone tools tree and into `gitgalaxy/recorders/sbom_recorder.py`, as a passive Phase 12 exit stage, so the manifest is tied to the same finalized in-RAM state as the forensic audits. It updated action.yml, smoke-test.yml and gitgalaxy/tools/README.md accordingly, but missed [project.scripts].

So this removes the entry rather than recreating the module: `galaxyscope <target> --sbom-only` is the supported path and re-adding a standalone binary would reverse the #83 decision. A comment records why, so the next reader doesn't "restore" it. Also fixes the three stale references left by the same incomplete migration:

- README.md linked to tree/main/gitgalaxy/tools/compliance/ (a 404) -> now points at recorders/sbom_recorder.py
- gitgalaxy/tools/README.md still listed zero-trust-sbom as a GitHub Action `tool:` option, though action.yml's own option list dropped it in #83
- the SBOM cookbook documented `python gitgalaxy/tools/sbom_generator.py`, a third dead path, and an output filename the engine never emits

Verified in a clean venv: all 19 remaining entry points import and expose a callable main() (the full-table sweep the issue asked for -- this was the only broken one), and `galaxyscope . --sbom-only` emits valid CycloneDX 1.4 with correct purls and populated gitgalaxy:trust_status properties. No Python source changed, so the parsing-logic Differential Scan does not apply.

Closes #1179
### Description of Structural Changes

None. This is a packaging and documentation fix — no Python source is touched, so no
structural signatures, regexes, or pipeline stages change. The diff is `pyproject.toml`
(one entry removed, one explanatory comment added) plus three markdown files.

### Visual Observatory Testing
- [ ] ~~I tested the output JSON on **GitGalaxy.io** OR the **Airgap Observatory**.~~
- [ ] ~~Flexbox constraints, HUD elements, and 3D rendering remain intact.~~

**N/A** — this PR cannot affect visualization output. No engine code changes, so the
`*_galaxy_gpu.json` payload is byte-identical to `main`. Marking these unchecked rather
than ticking them, since I did not run the Observatory and nothing here would have
changed what it renders.

### The Differential Scan Acknowledgement
- [x] I understand that my PR will be subjected to a Full Differential Scan.
- [ ] ~~I have provided the link to the specific repository this PR addresses so it can be tested alongside the 80-repo calibrated baseline.~~ **N/A** — this is not a parsing-accuracy fix against a specific repo; there is no target repository to add to the baseline.
- [x] I believe these changes will measurably improve the engine's Accuracy, Speed, Utility, or Ethos without causing regressions.
  - **Utility:** a documented-but-broken command is gone, and the three dead paths users were being pointed at now point at the real one.
  - **No regression risk:** the golden-master snapshots are unreachable from this diff, and the ruff/mypy baselines are unaffected (no `.py` files changed).

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,7 @@ An analytical suite for sanitizing mainframe monoliths. It safely neutralizes le
 ### [Software Supply Chain Security & Pre-Commit Firewalls](https://github.com/squid-protocol/gitgalaxy/tree/main/gitgalaxy/tools/supply_chain_security/)
 Pre-commit firewalls that scan physical file internals rather than trusting manifest files — built to block steganography, byte-level XOR decryption loops, homoglyph typosquatting, and exposed cryptographic vaults before they enter your CI/CD pipeline. **[Deploy directly via our GitHub Action](https://github.com/squid-protocol/gitgalaxy/blob/main/github-action-readme.md).**
 
-### [SBOM Generation & Dependency Auditing](https://github.com/squid-protocol/gitgalaxy/tree/main/gitgalaxy/tools/compliance/)
+### [SBOM Generation & Dependency Auditing](https://github.com/squid-protocol/gitgalaxy/blob/main/gitgalaxy/recorders/sbom_recorder.py)
 A Software Bill of Materials (SBOM) generator that doesn't blindly trust `package.json` or `requirements.txt` — it locates the physical dependencies on disk, checks their entropy and linguistic identity against what a legitimate version should look like, and generates strict CycloneDX 1.4 JSON reports.
 * **Benchmark:** Mapped and verified the physical internals of 170 unique Go modules inside the local Kubernetes repository. A single-repo result, not a claim of coverage across the Go ecosystem.
 
--- a/docs/wiki/cookbook/generate-zero-trust-sbom.md
+++ b/docs/wiki/cookbook/generate-zero-trust-sbom.md
@@ -11,10 +11,11 @@ GitGalaxy shifts this paradigm using the **Universal Zero-Trust SBOM Generator**
 The GitGalaxy generator supports NPM, Packagist (PHP), PyPI, and Cargo ecosystems. It performs a multi-stage physical audit to map the delta between what is declared and what actually exists in memory.
 
 ### 1. Execute the Generation
-Point the generator at the root directory of your project.
+Point the generator at the root directory of your project. SBOM generation is a native
+exclusive mode of the main orchestrator — there is no separate SBOM binary to install.
 
 ```bash
-python gitgalaxy/tools/sbom_generator.py /path/to/target_project
+galaxyscope /path/to/target_project --sbom-only
 ```
 
 ### 2. The Physical Integrity Audit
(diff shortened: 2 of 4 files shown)


squid-protocol (approved):

squid-protocol:
@ringoliRob Thanks for your help! It's always good to have a well maintained codebase. I appreciate your time.

Issue #1228, opened 2026-08-10
Build generalized tests/tools/tree_sitter_accuracy_audit.py + per-language node-map registry


squid-protocol:
Part of the tree-sitter accuracy epic (#1227).

## What to build

`tests/tools/tree_sitter_accuracy_audit.py`, mirroring `tests/ast_accuracy_audit.py`'s shape
(same CLI: no-args full report, `--ci` terse regression gate, `--regenerate` to bless a new
baseline) but generalized to take a `--lang <name>` argument and swap the ground-truth engine:

- **Corpus**: resolve `language-crucible` the same way `tests/tools/crucible_check.py` already
  does (`LANGUAGE_CRUCIBLE_PATH` env var, falling back to a sibling `../language-crucible`
  checkout) and scan `data/<lang>/` for that language's source files. Pin to the same tag
  `crucible_check.py`/`update_golden_master.py` use -- don't introduce a second, independently-
  drifting pin for the same corpus.
- **Engine side**: run `galaxyscope --db-only` against `language-crucible/data/<lang>/` (or reuse
  an existing `--db-only` scan if one's already cached for that run -- check how
  `tests/tools/self_scan.py`/`crucible_check.py` avoid redundant scans before writing a new
  scanning path from scratch).
- **Ground-truth side**: parse each file with `tree_sitter_language_pack.get_parser(<ts_lang>)`
  and walk the tree per a **per-language node-map config** -- a small dict per language giving:
  the tree-sitter grammar name if it differs from GitGalaxy's own language key (e.g.
  `shell`->`bash`), the function-like node type(s), the class-like node type(s), the field name
  holding the identifier, and the field name (or child-counting method) for parameter count. Start
  with just `javascript` (see below) and `python`... no -- Python already has `ast_accuracy_audit.py`
  and doesn't need tree-sitter; start with `javascript` only, matching an existing entry to
  validate against, then leave the registry easy to extend one language at a time.
- **Diff/metrics**: same shape as `ast_accuracy_audit.py` -- found/extra functions and classes,
  args-count exact-match rate, verbose sample output of mismatches. Reuse its `_regressions`/
  `_ground_truth_drift`/report-printing logic directly if the shapes line up closely enough to
  share code (a shared base module) rather than copy-pasting the whole file -- but don't force a
  shared abstraction if Python's `ast`-specific pieces (its `_real_arg_count`, its class-of-node
[...]

Issue #1227, opened 2026-08-10
Epic: Real-world tree-sitter/AST accuracy tracking, per language


squid-protocol:
## Why

`docs/language_status/<lang>.md` §9 ("Measured accuracy") is the strongest evidence this repo
has for whether `func_start`/`args`/`class_start`/`_dependency_capture` actually work on real,
unmodified production code -- as opposed to the (also necessary, but isolated) hand-picked
valid/invalid/pathological snippets in `tests/extraction/languages/test_<lang>.py`. It's the
methodology that surfaced #1182/#1183/#1184/#1193 for Python (via stdlib `ast`) and #1220/#1221/
#1222 for JavaScript (via `tree-sitter-language-pack`, the first non-Python repeat).

Both of those passes were one-off, hand-run investigations -- the JavaScript one explicitly used
an ad hoc scratchpad script, not committed tooling (see PR #1223's own description: "no committed
permanent tooling ... since the user chose 'single deep dive' not 'build the tooling'"). That
means every future language pass currently re-derives the methodology from scratch, and there's
no regression floor once a pass is done -- a later PR could silently regress a language's
real-world accuracy with nothing catching it (unlike `ruff_audit.py`/`mypy_audit.py`/
`dead_key_audit.py`/`tests/ast_accuracy_audit.py`, which are all baseline-gated).

This epic is the "build it once, then roll out per language" pass for tree-sitter-based accuracy
measurement, following the same shape as epic #813 (extraction gauntlets) and epic #1069
(strict-signature depth): one prerequisite tool-build issue, then one issue per language.

## Design (from investigation, see sub-issue #1 for the full writeup)

- **One generalized script**, `tests/tools/tree_sitter_accuracy_audit.py`, mirroring
  `tests/ast_accuracy_audit.py`'s proven shape (pinned corpus, `--ci`/`--regenerate`/full-report
  modes, per-language baseline JSON, regression-gated not zero-tolerance) -- swap `ast` for
  `tree-sitter-language-pack` and use a small per-language "node map" (function/class node types,
  name/params field names -- these genuinely vary per grammar, e.g. JS arrow functions use field
  `parameter` singular for a bare param but `parameters` plural for a parenthesized list) as the
  only hand-built per-language artifact.
- **Corpus: reuse `language-crucible`** (the same ~80-repo corpus `tests/tools/crucible_check.py`
[...]

PR #1231, opened 2026-08-10
docs: add SCOPE/LIMITATIONS and BASELINE sections to tree_sitter_accuracy_audit.py

squid-protocol:
Follow-up to #1229 (merged) — my independent-verification pass on that PR (before it merged)
found two real, non-blocking documentation gaps and pushed a fix commit to the same branch, but
the PR was merged before that push landed, so it never made it into `main`. This PR lands just
that one commit.

## What's added

A SCOPE & LIMITATIONS section and a BASELINE section to
`tests/tools/tree_sitter_accuracy_audit.py`'s docstring, matching `ast_accuracy_audit.py`'s own
documentation shape (which the tool's own design doc, #1228, asked it to mirror).

Documents two measurement-noise sources confirmed during verification, neither a functional bug:

1. **Same-name collision within a file**: a function is "found" by exact name match within its
   file (matching `ast_accuracy_audit.py`'s own documented trade-off) — not disambiguated by
   class/scope. Confirmed on the corpus: `jquery/event.js` defines two different functions both
   named `on` (a module-level 6-arg helper, and a 4-arg `.on()` prototype method) — the reported
   args-count "mismatch" for `on` in that file is this artifact, not a real args-counting bug in
   either GitGalaxy or the tree-sitter walker.
2. **Ground-truth gap on Flow-typed syntax**: tree-sitter's plain `javascript` grammar can't fully
   parse Flow type annotations (`function f(x: ?any): ?Iterator<any> {`), used in several
   `language-crucible/data/javascript/react/*.js` files. A Flow-typed function tree-sitter fails
   to recognize is invisible to `real_functions` (the ground truth), so if GitGalaxy's own
   grammar-agnostic regex still correctly matches it, it gets counted as a false "extra"
   (phantom) against GitGalaxy — an artifact of the measurement's ground-truth parser, not
   necessarily a real precision defect. Confirmed example:
   `react/ReactSymbols.js::getIteratorFn`.

No functional/behavior change — verified the tool's own numbers are identical before and after
this commit (`--ci` mode: "OK -- matches the committed baseline, no regressions"), and
`audit_check.py` + the full test suite (6815 passed) both stay clean.

Part of the tree-sitter accuracy epic (#1227).

### DIFF
--- a/tests/tools/tree_sitter_accuracy_audit.py
+++ b/tests/tools/tree_sitter_accuracy_audit.py
@@ -16,7 +16,54 @@ CORPUS
     Unlike ast_accuracy_audit.py which pins this repo's own code via git archive,
     this tool uses the external `language-crucible` corpus. By default, it expects
     it checked out as a sibling directory `../language-crucible`, or set via
-    `LANGUAGE_CRUCIBLE_PATH` env var.
+    `LANGUAGE_CRUCIBLE_PATH` env var. NOTE: "sibling directory" is relative to
+    THIS repo's own checkout root, not a git worktree's -- running from a worktree
+    (e.g. `gitgalaxy-worktrees/some-branch/`) needs `LANGUAGE_CRUCIBLE_PATH` set
+    explicitly, same pre-existing limitation `crucible_check.py` already has.
+
+BASELINE
+    tests/tree_sitter_accuracy_baseline_<lang>.json, one file per language (not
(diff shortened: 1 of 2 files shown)


squid-protocol:
Closing in favor of a clean-diff PR from a fresh branch -- this one's underlying patch was correct but GitHub was rendering it as a full-file diff (470 additions) because the branch's earlier commit predates the squash-merge on main, confusing for review. Same change, cleaner presentation.

PR #1543, opened 2026-08-14
docs: auto-update tree-sitter accuracy data

squid-protocol:
Automated update from the latest tree-sitter accuracy measurement on `main`.

## Notable changes since the previous batch
- **tcl** Args Exact-Match improved 16.2pp (83.8% → 100.0%)
- **dart** Func Precision improved 6.4pp (65.7% → 72.1%)

- Appends one row per language to `docs/self_scan/tree_sitter_accuracy_history.csv`.
- Regenerates the summary table in `gitgalaxy/standards/language_standards.py`.
- Re-renders `docs/self_scan/tree_sitter_accuracy_chart.svg` from that same batch.

Never touches `tests/tree_sitter_accuracy_baseline_*.json` -- those stay
human-reviewed via `tree_sitter_accuracy_audit.py --regenerate`.


squid-protocol:
Closing as stale/conflicted -- this scan ran against dart's merge (#1538) commit before tcl's fix (#1539) landed a minute later, so its own numbers don't include the tcl improvement its own summary claims, and it now conflicts with main's own auto-update (1252180d) which has the same staleness. Rather than hand-resolve generated-file conflicts, letting the next merge-triggered run regenerate this fresh and non-conflicting once the current batch of sweep PRs (#1538/#1539 landed, csharp-1427/python-1527/golden-master-bless still in flight) settles.

Issue #1452, opened 2026-08-13
javascript func_start: bare call-site identifiers with no capturing group fall back to whole-match-text naming, producing phantom functions


squid-protocol:
## Summary

`tests/tools/tree_sitter_accuracy_audit.py --lang javascript` measures func_start precision at
73.5%. One confirmed, reproducible false-positive mechanism: a bare call-site identifier (an
imported function being CALLED, not defined) is matched by some alternative in the `func_start`
regex with NO capturing group of its own, causing `detector.py`'s fallback naming logic
(`raw_name = match.group(match.lastindex) if match.lastindex else match.group(0)`, around lines
1809/2263/2383) to fall back to the ENTIRE match text (since `match.lastindex` is `None` when no
group participated) and extract a phantom function name from it.

## Repro

`language-crucible/data/javascript/jquery/css.js:314`:
```javascript
					swap( elem, cssShow, function() {
```
`swap` here is an imported function (`import { swap } from "./css/var/swap.js";`, line 12) being
CALLED with an inline anonymous callback -- not a definition of anything named `swap`. Confirmed
via direct `func_start.finditer()`: this line produces a match with `m.groups()` entirely
`(None, None, ...)` (no named group captured) and `m.group(0)` equal to `'\t\t\t\t\tswap'` (just
leading whitespace + the bare word `swap`, no parens/args/brace included in the match at all).
Since no group participated, `match.lastindex` is `None`, and the pipeline's fallback logic uses
the whole match text as the raw name, which `_extract_name` then reduces to `swap` -- reported as
a phantom function. Same shape recurs across the corpus for other imported-then-called names
(`attemptEarlyBailoutIfNoScheduledUpdate`, `bailoutOnAlreadyFinishedWork`, `beginWork` in
`react/ReactFiberBeginWork.js`, `addHelper` in `threejs/Editor.js`).

## Suggested fix shape

This looks like an alternative within javascript's `func_start` regex that's intended to detect
some OTHER construct (likely an anonymous-function/callback-adjacent shape, or a keyword-based
alternative without its own name-capture group) but is also incidentally matching a bare
call-site identifier with nothing to actually distinguish "this is a real anchor" from "this is
just an identifier token sitting at a matched position." Needs direct tracing of javascript's
`func_start` regex (`gitgalaxy/standards/language_standards.py`, `"javascript"` entry) to find
which alternative produces a match with zero capturing groups for input like `swap( elem, ...,
[...]

PR #1639, opened 2026-08-15
fix(typescript): match constructors/methods with callback-typed parameters

### LINKED ISSUE #1630: typescript func_start: nested parens in callback-typed parameters break the parameter-list terminator, dropping constructors/methods entirely
## Summary

Continuing the `typescript` func_start diagnosis from #1629 (same 159-miss diff, same corpus): a second, distinct root cause accounts for the rest of the recall gap not covered by #1629's brace-less-body drop — `func_start`'s parameter-list terminator breaks on a nested `(` inside a parameter's own type annotation (i.e. any callback-typed parameter), causing the **whole signature** to go unmatched at the regex level, not just misrecorded.

## Root cause

`gitgalaxy/standards/language_standards.py`'s `typescript` `func_start` rule's zero-prefix branch (the one that matches bare method names like `constructor(...)`) closes its parameter list with:

```
\([^)]*\)[^=;{]*=>[ \t\n]{0,50})?\{|:[^{;]{0,200}[ \t\n]{0,50};))
```

i.e. `\([^)]*\)` — a flat, non-nesting character class. If any parameter's own type annotation contains a `)` (which any callback/arrow-typed parameter does — `(x: () => void)` has an inner `()`), the class stops at the *first* `)` it sees, which is the inner callback's, not the real closing paren of the outer parameter list. The match then fails to find its terminator and the entire function/method/constructor signature is dropped — this is a **regex-level non-match**, confirmed directly (not a downstream `_slice_by_braces` issue like #1629):

```pycon
>>> rule.search("constructor(a: number) {")
<re.Match ...>  # matches fine
>>> rule.search("constructor(transport: ConnectionTransport, onDisconnect: () => void) {")
None            # nested `()` in the 2nd param breaks it
>>> rule.search("addObjectListener(eventName: (string | symbol), handler: (...args: any[]) => void) {")
None            # same shape, non-constructor method
```

This isn't a rare shape — callback-typed parameters (`onClose: () => void`, `handler: (err: Error) => void`, event-emitter-style constructors) are idiomatic TypeScript, especially in class constructors that wire up dependency callbacks.

## Evidence

19 of the 159 `found_functions` misses are named exactly `constructor` and land in files that follow this pattern, e.g. `language-crucible/data/typescript/playwright/bidiConnection.ts:48`:

```ts
constructor(transport: ConnectionTransport, onDisconnect: () => void, protocolLogger: ProtocolLogger, browserLogsCollector: RecentLogsCollector) {
```

Verified via a single-file `galaxyscope --db-only` scan of `bidiConnection.ts`: `function_data` has zero rows named `constructor` for the whole file (both classes' constructors, lines 48 and 169, missing), while every other real method in the file (`nextMessageId`, `rawSend`, `send`, etc. — none with a nested-paren param) is found correctly.

The same shape also drops non-constructor methods, not just constructors — e.g. `language-crucible/data/typescript/playwright/dispatcher.ts:89`:

```ts
addObjectListener(eventName: (string | symbol), handler: (...args: any[]) => void) {
```

## Suggested fix direction

Same class of fix already applied elsewhere in this same regex for balanced-generic step-overs (`(?:[^<>]|<[^<>]*>)*` tolerates one level of `<>` nesting) — extend the parameter-list class to tolerate one level of nested parens: `\((?:[^()]|\([^()]*\))*\)`, bounded the same way the generic step-over is, to avoid reintroducing a ReDoS shape (Rule 11 in `how_to_add_a_language.md`).

Filed under epic #1261, alongside #1629 (same corpus/diagnosis pass, different root cause).

uuzzrm:
## Summary

Fixes #1630

The zero-prefix branch of the TypeScript `func_start` rule closes the parameter list with a flat `\([^)]*\)` class, which stops at the first inner `)` it sees. Any callback-typed parameter carries an inner `()`, so for something like:

```ts
constructor(transport: ConnectionTransport, onDisconnect: () => void) {
```

the terminator lookahead failed at the first `)` of `()`, the `{` anchor was never reached, and the **whole signature silently stopped matching** at the regex level -- not a misrecord, a non-match. Constructors and methods with callback params were simply invisible to extraction.

Replaced the flat class with the same bounded one-level-nesting form the generic step-over already uses in this file:

```regex
\((?:[^()]|\([^()]*\))*\)
```

The two alternatives never match overlapping text, so it stays linear (same Rule 11 shape as `(?:[^<>]|<[^<>]*>)*`); no new ReDoS surface.

A nice side effect: the old flat class also made `describe('suite', () => {` (a bare call with an inline callback) match as if it were a method signature. With the parameter list now requiring the outer paren to actually close, that false positive is gone too -- mirroring what the JavaScript rule already does since #1452.

## Type of change

- [x] Bug fix
- [x] Parsing or engine logic (`gitgalaxy/core/detector.py`, `language_standards.py`, `prism.py`, or a per-language rule)
- [ ] Other (describe above)

## CI checklist

- [ ] `python tests/tools/audit_check.py` passes (bundles the ruff/mypy/dead-key baseline-gated audits + `ruff format --check`)
- [x] `python -m pytest tests/` passes, or the specific test file(s) for this change (list them below): `tests/extraction/languages/test_typescript.py` (81 passed), `tests/extraction/languages/test_javascript.py` (53 passed)
- [ ] **Touches parsing/engine logic (see above)?** `python tests/tools/crucible_check.py` passes -- this is the Differential Scan against the ~80-repo calibrated baseline (see `CONTRIBUTING.md`)
- [ ] **Changes the CLI's output JSON schema?** Verified on [GitGalaxy.io](https://gitgalaxy.io) or the local Airgap Observatory that 3D rendering still works

## Differential Scan target

Typescript parser corpus (playwright, fp-ts, vscode samples) -- the same files the issue's measurements came from.

## Verification

```console
$ uv run python -m pytest tests/extraction/languages/test_typescript.py -q
81 passed in 0.42s

$ uv run python tests/tools/tree_sitter_accuracy_audit.py --lang typescript
metric                 baseline    current
found_functions            1843       1876
extra_functions             309        299
args_exact_match           1779       1796
...
```

Found-functions up (1843 -> 1876), extra-functions down (309 -> 299), and the typescript accuracy baseline + summary table updated to match. Full `pytest tests/` run: 6913 passed; the only failures are pre-existing environment issues on this Windows box (missing xgboost for the ML auditor tests, symlink privileges, and a GBK console encoding crash in the golden-crucible harness) that also fail on a clean checkout here -- the golden-crucible check itself is CI's job on Linux.

Note on the golden master: the typescript `function_data` rows shift (more real functions now found, fewer phantom bare-call matches), so `tests/golden_master_audit.json` / `tests/golden_master_zero_dep_audit.json` will need re-blessing via `tests/tools/update_golden_master.py` on a clean checkout. I regenerated the zero-dep fixture locally, but this machine's corpus clone is shallow, which trips the harness's temporal-collapse neutralization and shifts coordinate floats unrelated to this change -- blessing that would bake environment noise into the baseline. Leaving the fixtures untouched for the Differential Scan to flag the real typescript-only delta.


uuzzrm:
CI status for reviewers: all lint/type/test gates are green (ruff, mypy, full-suite, smoke tests, tree-sitter accuracy audit). The two crucible-audit checks are failing because this PR changes TypeScript function detection -- the golden-master fixtures still carry the pre-fix counts (e.g. vscode/instantiationService.ts now finds the 8-param constructor that was previously dropped). The fixtures need re-blessing via tests/tools/update_golden_master.py on a clean checkout, which I deliberately didn't do from this machine (shallow corpus clone trips the temporal-collapse neutralization and would bake unrelated float noise into the baseline). Happy to push the re-blessed fixtures if you'd prefer I do it.

squid-protocol (approved):

uuzzrm:
Conflict with main is resolved and the golden-master fixtures were re-blessed against the merged tree, including the zig detection changes from #1647. Both crucible modes, the tree-sitter audit, full-suite, and smoke tests are green on the current head — ready to merge whenever convenient.

Issue #1631, opened 2026-08-14
typescript func_start: interface/type function-typed member declarations misidentified as real function definitions (majority of 309 extra_functions)


squid-protocol:
## Summary

`tests/tools/tree_sitter_accuracy_audit.py --lang typescript` measures 309 `extra_functions` (false positives) against `language-crucible/data/typescript`, dragging func_precision to 85.6%. Full diff (not just the tool's 8-file sample) shows the dominant driver: `func_start`'s colon-annotated-arrow branch cannot distinguish a real arrow-function *value* assignment from an `interface`/`type` member declared with a *function-type annotation* — both share the identical `IDENT: (params) => ReturnType` surface syntax, but only one has an actual implementation. GitGalaxy reports a phantom function for every function-typed interface/type field.

## Root cause

`gitgalaxy/standards/language_standards.py`'s `typescript` `func_start` rule's third alternative:

```python
r"^[ \t]*[a-zA-Z_$][\w$]*(?=[ \t\n]*:[ \t\n]*(?:async\s*)?(?:<(?:[^<>]|<[^<>]*>)*>\s*)?(?:function(?:\s*\*)?\b|\([^)]*\)[^=;{]*=>|[a-zA-Z_$][\w$]*[ \t\n]*=>))|"
```

matches any `IDENT:` followed eventually by `(...) => ...`, with no requirement that an actual `=` assignment or `{` body ever follows. That's correct for an object-literal method-shorthand-style property (`{ foo: (x) => x + 1 }`, a real function), but the exact same shape also describes an `interface`/`type` member's function-type annotation, which is pure type-level syntax with no runtime function at all:

```ts
export interface PipeableApply2<F extends URIS2> extends PipeableFunctor2<F> {
  readonly chainFirst: <S, R, E, A, B>(
    f: (a: A) => Kind4<F, S, R, E, B>
  ) => (ma: Kind4<F, S, R, E, A>) => Kind4<F, S, R, E, A>
}
```

Here `f: (a: A) => Kind4<F, S, R, E, B>` is a *parameter's* type annotation, one level nested inside `chainFirst`'s own function-type signature — not a definition. GitGalaxy reports a phantom function named `f`.

Confirmed directly against the regex:

```pycon
>>> rule.finditer('''export interface Foo {
...   readonly chainFirst: <S, R, E, A, B>(
...     f: (a: A) => Kind4<F, S, R, E, B>
[...]

PR #1656, opened 2026-08-15
fix(typescript): record expression-bodied arrow functions (#1629)

### LINKED ISSUE #1629: typescript func_start: brace-less/expression-bodied function definitions (curried FP-style consts) are silently dropped entirely
## Summary

`tests/tools/tree_sitter_accuracy_audit.py --lang typescript` currently measures func_recall=92.1%, extra_functions=309 (precision 85.6%) against the `language-crucible/data/typescript` corpus (see `docs/self_scan/tree_sitter_accuracy_history.csv`). Diagnosed the 159 `found_functions` misses directly (full diff dumped via a one-off script reusing `tree_sitter_accuracy_audit.py`'s own `measure()` internals, not just the tool's 8-file/3-name sample). This is the single largest contributor: **at least 88 of the 159 missing functions (>55%)** are brace-less, expression-bodied arrow function definitions — real, `func_start`-regex-matched function declarations that `detector.py`'s `_slice_by_braces` (Mode B) silently drops because their body has no `{` at all.

## Root cause

`_slice_by_braces` (`gitgalaxy/core/detector.py:2133`) has language-gated special handling for brace-less/expression bodies in `csharp` (#789, `=>` arrow-bodied members), `scala` (#1266, bare `= expr` bodies), `rust`/`zig`/`solidity` (bodyless `;`-terminated trait signatures), `kotlin`, `objective-c`, and `dart` — but **not** `typescript` (or `javascript`, which shares the same fallback). Every language not explicitly gated into one of those branches falls through to the generic default:

```python
else:
    brace_idx = safe_code.find(opener, start_idx, search_limit)
    if brace_idx == -1:
        continue
    end_idx = self._find_balanced_end(safe_code, brace_idx, opener, closer)
```

If no `{` appears within `search_limit` (min of 2000 chars or the next `func_start` match), the match is dropped entirely via `continue` — the function is never recorded, not even with a degraded/zero-magnitude entry.

TypeScript/JavaScript idiomatically use brace-less arrow function bodies extremely often — single-expression arrow functions (`const swap = (x) => x + 1`) and, especially in functional-style code, curried multi-arrow chains with no `{` anywhere in the whole definition. This is not a rare style; it's the primary export shape throughout `fp-ts` (a top-20 npm functional-programming library, part of the pinned corpus).

## Evidence

Confirmed with the actual `func_start` regex (`gitgalaxy/standards/language_standards.py`) — these DO match at the regex level (proving the drop happens in `_slice_by_braces`, not the regex):

```pycon
>>> rule.search("export const swap = <E, A>(ma: Either<E, A>): Either<A, E> => (isLeft(ma) ? right(ma.left) : left(ma.right))")
<re.Match object; span=(13, 17), match='swap'>   # matches, but absent from function_data
```

`language-crucible/data/typescript/fp-ts/Either.ts` alone accounts for dozens of these, e.g.:

```ts
// line 1325 -- matches func_start, never reaches function_data
export const swap = <E, A>(ma: Either<E, A>): Either<A, E> => (isLeft(ma) ? right(ma.left) : left(ma.right))

// line 688 -- curried chain, no `{` anywhere in the whole definition
export const traverse: PipeableTraverse2<URI> =
  <F>(F: ApplicativeHKT<F>) =>
  <A, B>(f: (a: A) => HKT<F, B>) =>
  <E>(ta: Either<E, A>): HKT<F, Either<E, B>> =>
    isLeft(ta) ? F.of(left(ta.left)) : F.map<B, Either<E, B>>(f(ta.right), right)

// line 170 -- top-level, unexported, still a real function
const _map: Monad2<URI>['map'] = (fa, f) => pipe(fa, map(f))
```

Confirmed via `galaxyscope --db-only` against a single-file scan of `fp-ts/Either.ts`: `function_data` has zero rows named `traverse`, `swap`, `_map`, `_reduce`, `_ap`, `_reduceRight`, `reduce`, `foldMap`, `mapLeft`, `filter`, `ap`, `bimap`, `altW`, `concat`, `equals` — all real, tree-sitter-confirmed functions.

Other affected files in the same corpus: `fp-ts/TaskEither.ts`, `fp-ts/pipeable.ts` (same style throughout).

## Suggested fix direction

Add a `typescript`/`javascript` branch to `_slice_by_braces` mirroring csharp's (#789) or scala's (#1266) approach: if no `{` is found before `search_limit`, look for the nearest un-nested `=>` after the parameter list and bound the body by either the next top-level statement terminator or the next `func_start` match (scala's approach — `end_idx = next_match_start` — is probably the closer analogy here, since TS/JS arrow bodies have no reliable `;` terminator either, same as scala's `def`).

## Scope note

This audit only measures `typescript`; `javascript` shares the exact same `_slice_by_braces` fallback path and near-certainly has the same gap (unverified here — out of scope for this issue, worth a quick follow-up check once this is fixed).

Filed under epic #1261.

uuzzrm:
## What & why

Closes #1629.

Expression-bodied arrow functions — the dominant export shape in functional TypeScript (`export const swap = <E, A>(ma) => isLeft(ma) ? ... : ...`, no `{` anywhere in the body) — were silently dropped by `_slice_by_braces`'s generic brace-only fallback. The func_start regex matched them fine; the brace search found no `{` and threw the match away. Per the issue's diagnosis this was the single largest recall gap in the typescript corpus: >55% of the 159 missing functions.

### The fix (`gitgalaxy/core/detector.py`)

Adds a typescript/javascript branch to `_slice_by_braces` that bounds a brace-less arrow's body by its own terminating `;` (paren/bracket-aware, so a `;` inside a nested literal can't truncate it) or by the next func_start match when semicolons are omitted (ASI). Three guards keep it from trading recall for precision:

1. **Assignment form only.** Only `IDENT = ... =>` (the value-assignment shape) gets brace-less handling. The line-start member form (`IDENT: ... =>`) is where interface/type function-typed members match (`bar: (x) => void;` in an interface is a type signature, not a function — the #1631 extra-functions driver), so those keep falling through to the generic path and stay dropped. Distinguished by the first top-level `=` after the match (a bare `=` vs the `=>`'s own equals), stopping at a top-level `{`/`;`.
2. **Named function-expression alias skip.** `track.createInterpolant = function Named(...) { ... }` is already recorded under the function expression's own name by func_start's function branch; recording the property alias too double-counts one function under two names.
3. **Return-type position reject.** An identifier immediately preceded by `=>` is a type position, not a value: in fp-ts's curried signatures (`... => M = (M) => ...`) the `M` is the return type and the assignment that follows belongs to the enclosing function.

### Measured effect (tree-sitter accuracy audit, pinned corpus)

| metric | baseline | with fix |
| --- | --- | --- |
| typescript found_functions | 1876 | **1950** |
| typescript extra_functions | 299 | **297** |
| javascript | unchanged | unchanged |

Both directions improved; no regressions in either language.

## Fixture changes

- `tests/tree_sitter_accuracy_baseline_typescript.json`: re-blessed via the audit's `--regenerate` path (found 1876→1950, extra 299→297, args_exact_match 1796→1855). `tests/tree_sitter_accuracy_baseline_javascript.json` is untouched — js metrics don't change.
- `gitgalaxy/standards/language_standards.py`: accuracy summary table regenerated (Typescript recall 93.7%→97.4%, precision 86.3%→86.8%) so it matches the committed baselines.
- `tests/golden_master_audit.json` / `tests/golden_master_zero_dep_audit.json`: the drift from this fix is confined to typescript/javascript files. The fixtures were regenerated locally, then the handful of non-typescript values that differ only because a local (Windows) scan computes a different float layout than CI (php/jcl/plaintext Doc Umbrella scores, the xray anomaly count) were restored to their committed values so the diff carries exactly the intended change and nothing else.

## Test plan

- New `test_detector_js_ts_expression_body_arrows_recorded`: 11 cases covering curried fp-style arrows, ASI (no semicolons), nested object literals, brace-in-param-type, class-field arrows, async arrows, and unchanged brace-bodied functions.
- Updated `test_detector_csharp_expression_body_fallback_gated_to_csharp_only`: the control is now php (a language with no brace-less handling) since js/ts now own theirs.
- `pytest tests/core_engine/test_detector.py tests/extraction/languages/test_{typescript,javascript}{,_strict}.py` → 429 pass; `full-suite` CI job passes.
- `ruff_audit.py --ci`: no new findings (baseline regenerated for two pure line-shifts).
- `dead-key-audit`, `ast-accuracy-audit`, `mypy-audit` (CI): pass. The 6 mypy findings my local Windows run reports are all in files this PR doesn't touch (`galaxyscope.py` SIGALRM, yaml stubs) and don't appear on CI.


uuzzrm:
Closing this in favor of #1655 — same fix for #1629, and #1655 is green across the board (both crucible modes, tree-sitter accuracy audit, ruff). This branch’s golden-master fixtures don’t match its code, which is why the crucible is red here.

Issue #1564, opened 2026-08-14
haskell func_start: single-line `let name args = expr` local bindings are structurally invisible


squid-protocol:
## Summary
Haskell's `let name args = expr` local bindings declared inline in `do` blocks are not captured by the current `func_start` regex, causing systematic under-detection of a common Haskell idiom.

## Root Cause
The `func_start` regex in `gitgalaxy/standards/language_standards.py` for Haskell has a local-binding alternative (second `|`-separated branch) that matches indented `name args = expr` lines, but excludes lines via a negative lookahead for keywords including `let`: `(?!(?:case|class|data|default|deriving|do|else|foreign|if|import|in|infix|infixl|infixr|instance|let|mdo|module|newtype|of|then|type|where)\b)`.

This lookahead is anchored right after the indentation, so it's meant to stop `let` itself from being mistaken for a function name — but it has the unintended side effect of blocking the *entire line* whenever it starts with `let `, meaning a real local binding declared inline as `let name args = expr` is never matched at all. The regex never retries past the `let ` token to find the real name that follows.

## Verification
Tested the regex directly in a REPL against this snippet, which returns zero matches:

```haskell
convertWithOpts scriptingEngine opts = do
  let outputFile = fromMaybe "-" (optOutputFile opts)
  let defFlavor fmt = Format.FlavoredFormat fmt mempty
  let isWarning msg = messageVerbosity msg == WARNING
```

Confirmed real-world instances missed in the pinned `language-crucible/data/haskell/pandoc` corpus (all genuine `let name args = expr` one-liners):
- `defFlavor` (App.hs:152)
- `makeSandboxed` (App.hs:165)
- `isWarning` (App.hs:114)
- `isPandocCiteproc` (Filter.hs — pattern-matched two-clause form at similar shape)
- `filterWithPath` (Filter.hs:52) — `let filterWithPath constr = maybe missingPath (return . constr . T.unpack)`

## Scope Note
This is distinct from #1435 (typeclass method signature floor) and #1442 (typeclass instance methods / `where`-clause locals without `::`). This issue is specifically the `let`-inside-`do`-block one-liner shape.

**Important:** Fixing this is a `detector.py`/`language_standards.py` regex change and needs the Differential Scan protocol (`crucible_check.py`) before merging per CLAUDE.md.

## Related Work
- #1532 (open)
- #1442 (closed)
- #1435 (closed)
- #1312 (closed)

PR #1661, opened 2026-08-15
fix(haskell): let func_start/args match same-line where-clause bindings

### LINKED ISSUE #1615: haskell func_start: single-line `where name args = expr` local bindings are structurally invisible
## Summary
A `where`-introduced local binding whose first clause is written entirely on one line — `where name args = expr` — is never matched by `func_start`. This is the exact same shape #1564 already fixed for `let`, just not yet extended to `where`.

## Root Cause
`func_start`'s equation-anchored alternative (`gitgalaxy/standards/language_standards.py`, haskell) is:
```
^[ \t]+(?:let[ \t]+)?(?!(?:case|class|data|default|deriving|do|else|foreign|if|import|in|infix|infixl|infixr|instance|let|mdo|module|newtype|of|then|type|where)\b)([a-z_][a-zA-Z0-9_']*)(?=[ \t]+[^\s=][^\n=]*(?<![!<>/])=(?![=>]))
```
The `(?:let[ \t]+)?` optional skip added by #1564 lets a same-line `let name args = expr` anchor past the `let` token. `where` gets no equivalent skip — when a line reads `where name args = expr` (the binding's first/only clause written on the same physical line as the `where` keyword itself, rather than `where` alone with the binding starting on the next line), the negative lookahead's keyword exclusion sees `where` immediately after `^[ \t]+` and rejects the whole line, with no way to retry past it the way `let` now can.

## Verification
Tested the regex directly:
```python
from gitgalaxy.standards.language_standards import LANGUAGE_DEFINITIONS
fs = LANGUAGE_DEFINITIONS['haskell']['rules']['func_start']
snippet = """
renderTags' :: [Tag T.Text] -> T.Text
renderTags' = renderTagsOptions
               renderOptions{ optMinimize = matchTags ["hr"] }
              where matchTags tags = flip elem tags . T.toLower
"""
list(fs.finditer(snippet))  # -> only matches renderTags', never matchTags
```

Confirmed real-world misses in the pinned `language-crucible/data/haskell/pandoc` corpus (found via the clause-collapsed recall check in #1614 — these are genuine misses, not clause-counting artifacts):
- `matchTags` (Shared.hs:734) — `where matchTags tags = flip elem tags . T.toLower`
- `rejectBadYear` (Shared.hs:283) — `where rejectBadYear day = case toGregorian day of`
- `go` (Shared.hs:261, inside `tabFilter`) — `where go s =` (technically spans onto the next line for the body, but the `where go s =` opener is the same single-line-`where` shape; `go`'s *other* 6 occurrences elsewhere in the same file, which use `where go :: ...` on its own line, are all found correctly)

## Proposed Fix
Extend the existing `(?:let[ \t]+)?` optional skip to also accept `where`, e.g. `(?:(?:let|where)[ \t]+)?`, mirroring #1564's fix exactly.

**Important:** This is a `detector.py`/`language_standards.py` regex change and needs the Differential Scan protocol (`crucible_check.py`) before merging per CLAUDE.md.

## Related Work
- #1564 (closed) — identical fix shape, for `let`
- #1614 (audit-tool ground-truth clause-collapsing bug — this issue's misses were isolated using that investigation)
- #1442, #1435 (closed) — other haskell func_start local-binding gaps

squid-protocol:
## Summary

Fixes #1615.

A `where`-introduced local binding whose first clause is written entirely on
one line (`where name args = expr`) was never matched by `func_start`'s
equation-anchored alternative -- the same shape #1564 already fixed for
`let`, just not yet extended to `where`. `func_start`'s negative-lookahead
keyword-exclusion list sees `where` immediately after `^[ \t]+` and rejects
the whole line, with no way to retry past it the way `let` now can.

Extended the identical optional-skip treatment #1564 already gave `let`
(`(?:let[ \t]+)?` -> `(?:(?:let|where)[ \t]+)?`) to `func_start`'s equation-
anchored alternative. While verifying, found the sibling `args` rule (which
got its own symmetric `let`-skip from #1564) needed the identical `where`
extension too -- without it, the newly-found functions from `func_start`
would get a wrong 0-arg count instead of simply being absent as before
(measured: `args_exact_match` briefly regressed 129 -> 128 with only the
`func_start` fix; fixing `args` too brought it to 132, net improvement).

## Verification

```
$ venv/bin/pytest tests/extraction/languages/test_haskell.py tests/extraction/languages/test_haskell_strict.py -q
177 passed

$ tree_sitter_accuracy_audit.py --lang haskell
found_functions   139 -> 142   (recall up)
extra_functions     2 ->   2   (unchanged, no new false positives)
args_exact_match  129 -> 132   (up)

$ tree_sitter_accuracy_audit.py --all --ci
31 language(s) checked, all OK.
```

`crucible_check.py --mode both` shows the expected cascading drift (global
PageRank/spatial-coordinate stats shift slightly corpus-wide when any one
language's function count changes) -- blessed via `--update --yes`, re-ran
to confirm PASS/PASS. `audit_check.py --ci` clean (ruff/mypy/dead-key/ast-accuracy).

## Type of change

- [x] Bug fix
- [x] Parsing or engine logic (`language_standards.py`)

## Differential Scan target

Haskell parser corpus (pandoc) -- the same files the issue's measurements came from.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


squid-protocol:
Reopening to try to shake loose a CI run -- matches this repo's documented 'push events silently stop creating runs' flake (see gitgalaxy.yml's own comment on this).

PR #1664, opened 2026-08-15
fix(dart): exclude try/finally from func_start phantom-function matches

### LINKED ISSUE #1622: dart: `try`/`finally` keywords misdetected as zero-arg function declarations
## Summary

Dart's `try` and `finally` keywords at line start are misdetected as phantom zero-arg function declarations because the `func_start` regex's negative-lookahead exclusions (at alternatives A/B/C boundaries) exclude `class|mixin|enum|extension|typedef|if|for|while|switch|catch|case|when|assert|return|throw|new|var|final|const` — but never `try` or `finally`. So `try {` / `finally {` matches as a function literally named `"try"`/`"finally"`.

## Confirmed repro

Direct regex test (from activated venv with `gitgalaxy` installed):

```python
>>> from gitgalaxy.standards.language_standards import LANGUAGE_DEFINITIONS
>>> rule = LANGUAGE_DEFINITIONS["dart"]["rules"]["func_start"]
>>> rule.search("    try {")
<re.Match object; span=(0, 9), match='    try'>  # captures group 3 = 'try'
>>> rule.search("    finally {")
<re.Match object; span=(0, 13), match='    finally'>  # captures group 3 = 'finally'
```

Corpus impact (confirmed via `tests/tools/tree_sitter_accuracy_audit.py --lang dart`):
- `language-crucible/data/dart/flutter/*.dart` corpus: 38 rows in `function_data` literally named `"try"`
- These account for ~9% of dart's remaining `func_start` false-positive rate
- The same gap applies to `finally` (though this corpus has zero `finally`-named false positives)

## Fix direction

Add `try` and `finally` to the negative-lookahead exclusion lists in dart's `func_start` regex (all 3 occurrences at alternatives A/B/C boundaries), matching the existing `catch` handling immediately adjacent to them in the pattern.

## Labels

`bug`, `core-engine`

## File location

`gitgalaxy/standards/language_standards.py` line ~9322-9345 (dart `func_start` regex)

squid-protocol:
## Summary

Fixes #1622.

Dart's `func_start` negative-lookahead exclusion lists (6 occurrences across
the 3 alternative A/B/C boundaries -- each alternative gates both an optional
return-type prefix and the captured name) excluded
`class|mixin|enum|extension|typedef|if|for|while|switch|catch|case|when|
assert|return|throw|new|var|final|const`, but never `try` or `finally`, so
`try {` / `finally {` at line-start matched as a phantom zero-arg function
literally named `"try"`/`"finally"`. Added `try|finally` adjacent to the
existing `catch` handling in all 6 occurrences (the issue's own text said
"3 boundaries" -- each boundary turned out to have 2 copies of the list, one
guarding the optional return-type prefix, one guarding the captured name).

## Verification

```
$ venv/bin/pytest tests/extraction/languages/test_dart.py tests/extraction/languages/test_dart_strict.py -q
188 passed, 1 skipped, 5 xfailed

$ tree_sitter_accuracy_audit.py --lang dart
extra_functions   186 -> 149   (-37, matches the issue's ~38 estimate)
found_functions  1340 -> 1340  (unchanged, no recall loss)

$ tree_sitter_accuracy_audit.py --all --ci
31 language(s) checked, all OK.
```

`crucible_check.py --mode both` shows the expected cascading drift (global
PageRank/spatial-coordinate stats shift slightly corpus-wide when any one
language's function count changes) -- blessed via `--update --yes`. Merged
origin/main in along the way (picked up #1654/#1655's typescript fixes,
which also touch the golden masters) and re-blessed once on top of that.
`audit_check.py --ci` clean (ruff/mypy/dead-key/ast-accuracy).

## Type of change

- [x] Bug fix
- [x] Parsing or engine logic (`language_standards.py`)

## Differential Scan target

Dart parser corpus (flutter) -- the same files the issue's measurements came from.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


squid-protocol:
Reopening to try to shake loose a CI run -- matches this repo's documented 'push events silently stop creating runs' flake (see gitgalaxy.yml's own comment on this).

Issue #1335, opened 2026-08-12
objective-c args extraction: untyped keyword-message params undercount to 0, and body content can leak into the count


squid-protocol:
## What

While investigating #1314's objc `func_start` precision fix (PR #1328) and its bodyless-`@interface`-declaration follow-up (PR #1332), args-count verification against `language-crucible/data/objective-c/worldwideweb` surfaced two related, unfixed gaps in objc's args extraction. Both are real, confirmed-against-corpus GitGalaxy defects (not measurement-tool artifacts).

**1. Untyped keyword-message parameters undercount to 0.** GitGalaxy's args regex (`language_standards.py`'s objc `"args"` rule, branch 1) requires a `(Type)` cast between each `label:` and its parameter name. Older, still-valid Objective-C style omits the type (defaults to `id`) -- e.g. `- back:sender { ... }`, `- help:sender`, `+ setManager:aManager`. None of these are recognized as having any parameters at all (`args=0` instead of 1+). Confirmed via `HyperManager.m`: `back`, `help`, `save`, `open`, `traceOn`, `searchRTF`, and roughly 15 others in that one file alone.

**2. When a method's own signature doesn't match ANY args branch** (case 1 above, or a genuinely zero-arg method with no colon at all), **args extraction falls through past the real signature and matches the first C-style-shaped call/statement found later in the method's own body** -- because `args_pattern.search(block)` searches the whole sliced function body, not just the signature line. Confirmed concretely:
- `Anchor.m`'s `- free { if (Address) free(Address); ... }` (zero-arg, no colon) measures `args=1`, borrowed from the `free(Address);` call inside its own body.
- `HyperManager.m`'s `- closeOthers:sender { ... printf("...", w); ... }` (1 real untyped param) measures `args=2`, borrowed from the unrelated `printf(...)` call's 2 comma-separated arguments.

This is the same architectural shape as #1221's "Invocation Shield" (bare call statements false-positive-matching as function signatures), but for objc's **args** regex specifically rather than `func_start` -- #1221's fix didn't cover this pillar/language combination.

## Why this needs care, not a quick regex patch

[...]

PR #1671, opened 2026-08-15
fix(c): args regex missed lowercase-typedef first params (issue #1646)

### LINKED ISSUE #1646: c args: RETURN_IF_ERROR(...) call inside function body matched as the signature's own parameter list
## Summary

Found incidentally while root-causing #1645 (zig args accuracy).
`cpython/compile.c::_PyCompile_MaybeAddStaticAttributeToClass` has a real
2-parameter signature:

```c
int
_PyCompile_MaybeAddStaticAttributeToClass(compiler *c, expr_ty e)
{
    assert(e->kind == Attribute_kind);
    ...
    RETURN_IF_ERROR(PySet_Add(u->u_static_attributes, e->v.Attribute.attr));
```

but the `args` regex's match for this function is NOT the signature at all --
it's `(PySet_Add(u->u_static_attributes, e->v.Attribute.attr))`, a call
expression from several lines into the function body. Confirmed by adding a
temporary debug print in `detector.py::_calculate_block_metrics` right after
`args_str` is captured and scanning just this one file with `galaxyscope`.

## Why this stayed hidden until now

Before #1645, `_count_top_level_args`'s depth counter treated a bare `>` as a
generic-closing bracket, including the `>` inside C's `->` pointer-member
operator. On this specific (wrong) matched text, that spurious decrement
happened to land the depth back to 0 right where the real top-level comma
sits, producing `args_count=2` -- which happens to equal the REAL signature's
arity by coincidence. #1645 correctly stopped treating `->`'s `>` as a
bracket for zig's own counting path, which would also fix this string if it
reached the same code path -- but since this instance goes through the
default (non-`treat_as_body`) wrapper-detection branch, which #1645
deliberately left unchanged (scoped fix, see that issue's description), it
still hits this pattern and now computes `args_count=1` from the same wrong
matched text: a different wrong answer than before, not a fix.

Net effect on the tree-sitter accuracy baseline: neutral (one already-broken
case flipped from "coincidentally right" to "correctly computed for the
wrong input" -- `tests/tree_sitter_accuracy_baseline_c.json` needed no
change, confirmed via `--regenerate` showing 0 net regressions once the
`=>`/`->` fix was scoped to skip this path).

## Real bug

The C `args` regex (or whatever slicing feeds it `args_search_text`) is
matching a call expression deep in the function body instead of the actual
`(compiler *c, expr_ty e)` signature. This is very likely a false-negative on
the signature match itself (the regex fails to match the real declaration
line for some reason specific to this signature's shape) causing `.search()`
to fall through to the next thing in `block` that superficially looks like
`identifier(...)`. Needs the same kind of investigation #1265/#1263 did for
c/cpp func_start: check whether the leading `int\n` return-type-on-its-own-
line shape, or something about `compiler *c` (pointer param with no space
before `*`), is defeating the current regex.

## Scope note

This is a distinct, pre-existing bug from #1645's zig fix -- not introduced
by it, just surfaced during that investigation. Filing per repo convention
("file incidental findings as issues" rather than folding out-of-scope fixes
into an unrelated PR).

uuzzrm:
Fixes #1646.

## Problem

The C `args` regex's typed-parameter whitelist only accepted primitives,
PascalCase types and `_t`-suffixed typedef names. A signature whose FIRST
parameter is a plain lowercase typedef (`compiler *c`, `expr_ty e`,
`mod_ty mod`) failed the whole regex, so `args_pattern.search()` on the
function block fell through to a lookalike `identifier(...)` call deeper
in the body -- e.g. `RETURN_IF_ERROR(PySet_Add(u->u_static_attributes,
e->v.Attribute.attr))` for
`_PyCompile_MaybeAddStaticAttributeToClass(compiler *c, expr_ty e)` -- and
counted the wrong text as the signature's parameter list.

## Fix

Added a `[a-z_]\w*` branch that requires the type+name pair (with optional
pointer stars between), so the signature matches while a bare call
argument (`helper(a, b)`) still can't pass as a "type". The signature sits
at the top of the block, so a genuine match always beats any body call.

## Verification

- `tests/tools/tree_sitter_accuracy_audit.py --lang c`:
  `args_exact_match` 1609 -> 1667 (+58), `found_functions`/`extra_functions`
  unchanged. The three remaining mismatches were verified pre-existing
  (same wrong answers as before the fix, just shifted match text).
- Baseline `tests/tree_sitter_accuracy_baseline_c.json` regenerated via the
  official tool (only `args_exact_match` changed).
- Regression tests added to `tests/extraction/languages/test_c.py` for the
  lowercase-typedef signature shape, the full issue repro (return type on
  its own line + body call), and negative cases (bare lowercase call args).
- Golden masters re-blessed with `tests/tools/update_golden_master.py`
  (both full-precision and zero-dependency modes), run on a Linux runner
  to match the golden-crucible CI environment: C files now report correct
  per-file `Function Parameters`/`Input Parameters` counts, which ripples
  into the corpus-wide aggregates the fixtures snapshot.

Signed-off-by: Ruiming Zhao <uuzzrm@gmail.com>

### DIFF
--- a/gitgalaxy/standards/language_standards.py
+++ b/gitgalaxy/standards/language_standards.py
@@ -3371,7 +3371,22 @@ LANGUAGE_DEFINITIONS: dict[str, Any] = {
                 # convention (`_PyStackRef`, `_Bool`, ...), extremely common in
                 # cpython internals and not covered by cpp's own version of this
                 # fallback.
-                r"(?!(?:if|for|while|switch|return|sizeof|typeof|_Alignof|__typeof__|__builtin_[a-zA-Z0-9_]+)\b)\b([a-zA-Z_]\w*)[ \t\n*]*(\(\s*(?:const\s+|volatile\s+)?(?:int|char|void|float|double|long|short|unsigned|signed|struct|enum|_*[A-Z]\w*|[a-z_]\w*_t)\b(?:[^)(]|\([^)]*\))*\))",
+                # BUG FIX (#1646): the whitelist still missed plain lowercase
+                # typedef names (no `_t` suffix) that are everywhere in real
+                # C internals -- `compiler *c`, `mod_ty`, `expr_ty`,
+                # `asdl_seq *`, `identifier`. A signature whose FIRST
+                # parameter uses one (_PyCompile_MaybeAddStaticAttributeToClass
+                # (compiler *c, expr_ty e)) failed the whole regex, and
+                # args_pattern.search() on the block fell through to a
+                # lookalike `identifier(...)` call deeper in the body
+                # (e.g. RETURN_IF_ERROR(PySet_Add(...))), counting the wrong
+                # text. Added a `[a-z_]\w*` branch requiring the type+name
+                # pair (`compiler *c`, `expr_ty e`, with optional pointer
+                # stars between) -- so the signature is accepted while a
+                # bare call argument (`helper(a, b)`) still can't pass as
+                # a "type". The signature sits at the top of the block, so
+                # a genuine match always beats any body call.
+                r"(?!(?:if|for|while|switch|return|sizeof|typeof|_Alignof|__typeof__|__builtin_[a-zA-Z0-9_]+)\b)\b([a-zA-Z_]\w*)[ \t\n*]*(\(\s*(?:const\s+|volatile\s+)?(?:int|char|void|float|double|long|short|unsigned|signed|struct|enum|_*[A-Z]\w*|[a-z_]\w*_t|[a-z_]\w*(?:\s*[*&]+\s*|\s+)[a-zA-Z_]\w*)\b(?:[^)(]|\([^)]*\))*\))",
                 re.M,
             ),
             # 3. linear (Sequential Boundaries)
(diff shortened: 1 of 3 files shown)


squid-protocol:
Thanks for this -- independently correct diagnosis (the `args` regex's type-alternation whitelist missing plain lowercase typedef names like `compiler`), and confirms the analysis.

This landed moments earlier via #1672 (merged), which took a slightly broader fix: in addition to widening the same type-alternation whitelist, it also bounds the args search to the signature's own text (`args_sig_end`, generalizing the existing #1335 objc mechanism) rather than widening the whitelist alone. The bounding half matters because a widened-but-still-incomplete whitelist can still fall through to a body call for some other type shape not covered by either whitelist -- bounding the search closes that whole class, not just this one repro.

Closing as superseded by #1672 to avoid a duplicate/conflicting fix to the same regex. Appreciate the quick turnaround -- this is exactly the kind of overlap that happens when two people pick up the same issue around the same time, not a quality issue with this PR.

uuzzrm:
No worries at all -- the search-bounding half is the right call, it closes the whole fall-through class instead of just this whitelist gap. Glad it's in. Thanks for the quick look!

PR #1673, opened 2026-08-15
fix(dart): stop treating try/finally blocks as function declarations (#1622)

### LINKED ISSUE #1622: dart: `try`/`finally` keywords misdetected as zero-arg function declarations

uuzzrm:
## What & why

Closes #1622.

Dart's `try` and `finally` keywords at line start were misdetected as phantom zero-arg function declarations: the `func_start` regex's negative-lookahead keyword exclusions covered `class|mixin|enum|extension|typedef|if|for|while|switch|catch|case|when|assert|return|throw|new|var|final|const` but never `try` or `finally` — so `try {` / `finally {` matched as a function literally named "try"/"finally". Confirmed via the issue's repro against `language-crucible/data/dart/flutter/*.dart`: 38 spurious "try" rows in `function_data`, ~9% of dart's remaining `func_start` false-positive rate.

### The fix

Add `try|finally` to the keyword-exclusion lookaheads in all three `func_start` alternatives (6 exclusion sites in `gitgalaxy/standards/language_standards.py`). A real method named `tryMethod` still matches; only the reserved keywords are rejected.

### Measured effect (tree-sitter accuracy audit, pinned corpus)

| metric | baseline | with fix |
| --- | --- | --- |
| dart found_functions | 1340 | **1340** (unchanged) |
| dart extra_functions | 186 | **149** (**-37** phantom try/finally functions removed) |

Zero recall loss, no regressions. Golden-master fixtures re-blessed via `update_golden_master.py` (both modes); the drift is confined to dart entries plus the usual sub-epsilon float drift from the shared spatial layout recomputing on the changed graph.

## Test plan

- 4 new invalid cases in `tests/extraction/languages/test_dart.py`: `try {`, `finally {`, indented variants.
- `pytest tests/extraction/languages/test_dart.py test_dart_strict.py` → 190 pass.
- `tree_sitter_accuracy_audit.py --lang dart` → no regressions (baseline re-blessed: extra 186→149).
- `ruff_audit.py --ci`: no new findings; `dead-key-audit`/`ast-accuracy-audit` clean.


squid-protocol:
Same fix, independently arrived at -- both add \`try|finally\` to the same 6 occurrences of the exclusion list. This is already up as #1664 (open, same diagnosis and fix shape), so closing this one as a duplicate to avoid two competing PRs on the same regex region. Nothing wrong with this PR -- just landed a few minutes behind.

Issue #1158, opened 2026-08-09
File-level K-Means archetype classification silently truncates against its own trained models (115-dim global vs. 83-dim live vector)


squid-protocol:
## 🛑 Problem Statement

`file_data.file_archetype` (`telemetry["archetype"]`) is assigned by `_classify_archetype()`
(`signal_processor.py:139-166`) against two separate pre-trained K-Means models, and both are out
of sync with the live feature vector actually fed into them.

**The live vector** (`signal_processor.py:609-633`, `raw_vector`) is built by iterating
`SIGNAL_SCHEMA` (skipping `indent_tabs`/`indent_spaces`/`hardware_bridge`/`cryptography`/`sec_*`
keys) and appending 7 engineered features (`cfr`, `log_logic_loc`, `log_imports_out`,
`log_popularity_in`, `log_max_func_comp`, `log_avg_func_args`, `log_churn`). Against the current
`SIGNAL_SCHEMA`, that's:

```
SIGNAL_SCHEMA length: 93
filtered length: 76
raw_vector length: 76 + 7 = 83
```

**Global model** -- `GENERAL_FILE_INFERENCE_MODEL` (`analysis_lens.py:2155`, `SCALER_MEDIANS` /
`SCALER_IQRS` / 18-cluster `ARCHETYPES_K18` centroids) is **115-dimensional**:

```
SCALER_MEDIANS (file) length: 115
GLOBAL_ARCHETYPES cluster count: 18
file_cluster_0 centroid len: 115
```

`_classify_archetype`'s distance loop (`signal_processor.py:156`) uses
`range(min(len(scaled_vector), len(centroid_vector)))`, so it silently compares only the first 83
of each centroid's 115 dimensions -- **32 trained dimensions (~28%) are dropped on every file.**

**Per-language ("local") models** -- `SPECIFIC_FILE_INFERENCE_MODEL` (`analysis_lens.py:4501`) --
are **74-dimensional** and only exist for **5 of the 59 languages** this engine parses (python,
javascript, typescript, c, cpp; confirmed via `LANGUAGE_DEFINITIONS`). Every other language's
files never get a `local_archetype` at all (`signal_processor.py:654`'s `lang_brain` lookup
misses), and even the 5 covered languages compare 74-of-83 live dimensions against their own
centroids -- same truncation bug, smaller magnitude.

**This also breaks a shipped, published claim.** `docs/wiki/03-05-claim-5-file-archetypes.md`
[...]

PR #1724, opened 2026-08-16
fix(metrics): guard archetype classification against silent dimension truncation (#1157)

### LINKED ISSUE #1157: Function-level K-Means archetype classification only compares 5 of the trained model's 62 dimensions
## 🛑 Problem Statement

`function_data.func_archetype` (and every function's in-RAM `telemetry["archetype"]`) is assigned
by a hand-rolled Euclidean nearest-centroid classifier in `signal_processor.py:561-587`,
independent of the shared `_classify_archetype()` helper the file-level classifier uses.

The live feature vector built for every function (`signal_processor.py:562-568`) has exactly
**5** dimensions:

```python
raw_vec = [
    float(s.get("branch", 0)),
    float(s.get("loc", 0)),
    float(s.get("args", 0)),
    float(s.get("keyword_density", 0.0)),
    float(s.get("control_flow_ratio", s.get("cf_ratio", 0.0))),
]
```

But the pre-trained model it's compared against -- `GENERAL_FUNCTION_INFERENCE_MODEL` in
`analysis_lens.py:1253` (`ARCHETYPES_K12`, 12 clusters) -- has **62**-dimensional
`SCALER_MEDIANS`, `SCALER_IQRS`, and centroid vectors. Confirmed by direct inspection:

```
func SCALER_MEDIANS len: 62
func arch key: ARCHETYPES_K12
func centroid count: 12
fxn_cluster_0 centroid len: 62
```

The distance loop at `signal_processor.py:578`:

```python
dist = math.sqrt(sum((a - b) ** 2 for a, b in zip(scaled_vec, centroid)))
```

uses `zip()`, which silently truncates to the shorter sequence -- so every function is classified
using only 5 of the 62 dimensions the model was actually trained on. **~92% of the trained
model's discriminating signal is discarded on every single classification.** This isn't a crash
or an empty column (`func_archetype` is populated on every function) -- it's a confidently-wrong
label, which is worse: nothing signals the classification is untrustworthy.

## 🎯 Objective

Restore function-archetype classification to actually use the dimensions the model was trained
on (or retrain/ship a model matched to the 5 live features, if 5 was in fact always the intended
scope -- unclear which is true, needs an owner decision, see Acceptance Criteria).

## 🛠️ Proposed Implementation

1. Decide the intended scope: either (a) the 5-feature live vector is a regression -- more
   features used to be computed and fed in before some past refactor -- and needs to be restored,
   or (b) the 62-dim model is legacy from a differently-scoped training run and needs to be
   retrained against today's 5-feature (or intentionally-expanded) vector. Check git history
   around `GENERAL_FUNCTION_INFERENCE_MODEL`'s introduction and the `raw_vec` construction for
   the answer.
2. Whichever direction is chosen, add a loud failure (not a silent `zip()` truncation) when a
   live feature vector's length doesn't match the model's trained dimensionality -- e.g.
   raise/log an error and fall back to `"Unclassified"` rather than quietly comparing a truncated
   vector. This is the same "silent schema drift" failure class as #101, #313, #1147, #325 -- this
   repo has hit it enough times that the guard is worth adding once, here, rather than
   re-discovering the next drift by accident.
3. While fixing this, consider replacing the hand-inlined distance loop with the shared
   `_classify_archetype()` helper (`signal_processor.py:139-166`) already used for file-level
   classification, so there's one implementation of "K-Means nearest-centroid classify" instead of
   two that can independently drift (echoes #249's "two disagreeing implementations" lesson).

## ✅ Acceptance Criteria

- [ ] Root cause of the 5-vs-62 mismatch identified and documented (regression vs. legacy model).
- [ ] Live feature vector and model dimensionality match, verified by a real test (e.g. assert
      `len(raw_vec) == len(GENERAL_FUNCTION_INFERENCE_MODEL["SCALER_MEDIANS"])`).
- [ ] A dimension mismatch between a live feature vector and its centroid model fails loudly
      instead of silently truncating via `zip()`.

uuzzrm:
## Summary

Fixes #1157, and covers the loud-failure acceptance criterion of #1158 (the shared `_classify_archetype()` guard applies to file-level classification too).

`func_archetype` (and `file_archetype`/`local_archetype`) are assigned by Euclidean nearest-centroid classification against pre-trained K-Means models that carry no feature-name metadata and have no training script in this repo. The live feature vectors — 5-dim per function, 83-dim per file — have never matched the shipped model dimensionalities (62-dim function, 115-dim global file, 74-dim per-language). The distance loops silently truncated to the shorter sequence via `zip()` / `min()`, so every function and file was labeled from a fragment of the model: confidently wrong, with nothing signaling the label was untrustworthy.

What this PR changes:
- `_classify_archetype()` now refuses to compare vectors of different lengths. It logs the mismatch once per `(vector_len, centroid_len, model)` pair and returns `Unclassified` instead of truncating — the loud-failure behavior both issues ask for.
- Function-level classification now routes through the same `_classify_archetype()` helper instead of a second hand-rolled distance loop, so the two classifiers can't drift apart. With the current stale model this leaves functions `Unclassified` (with a warning) rather than emitting truncated labels; as soon as the model is retrained against the live feature set, real labels come back automatically.
- Regression tests pin all of the above: mismatch → `Unclassified` + warning; matching dimensions → normal classification, for both the helper and the function/file paths.

## Type of change

- [x] Bug fix
- [ ] Parsing or engine logic (`gitgalaxy/core/detector.py`, `language_standards.py`, `prism.py`, or a per-language rule)
- [ ] Other (describe above)

## CI checklist

- [x] `python tests/tools/audit_check.py` passes — ruff (83-finding baseline, format clean), dead-key, ast-accuracy all OK. mypy on this Windows box reports only the pre-existing environment findings (SIGALRM absent on Windows, missing yaml stubs) that also appear on a clean `main` checkout; the ubuntu CI runner doesn't see them.
- [x] `python -m pytest tests/core_engine/test_signal_processor.py` — 62 passed
- [x] Full `python -m pytest tests/` — 6946 passed, 8 failed; the same 8 fail on a clean `main` checkout in this environment (Windows symlink privilege in one guidestar test, xgboost import in the security_auditor tests), so none are regressions from this change.
- [ ] Touches parsing/engine logic? No — `crucible_check.py` / the Differential Scan does not apply to a `metrics/` change.
- [ ] Changes the CLI's output JSON schema? No.

## Differential Scan target

N/A — not a parsing/engine change.

## Verification

```
$ python -m pytest tests/core_engine/test_signal_processor.py -q
62 passed

$ python tests/tools/audit_check.py
[ruff] OK -- 83-finding baseline, format clean.
[dead-key] OK
[ast-accuracy] OK -- matches the committed baseline, no regressions.

$ python -m pytest tests/ -q
6946 passed, 8 failed (same 8 pre-existing/environmental failures on clean main)
```

The `tests/ruff_audit_baseline.json` change is purely line-number shifts for the four pre-existing `signal_processor.py` findings moved by the added lines (regenerated via `audit_check.py --regenerate`).


squid-protocol (approved):

squid-protocol:
Thanks! @uuzzrm, could you please update the golden crucible resilience test? You rock!

Issue #1319, opened 2026-08-12
rust func_start misses bodyless multi-line trait method signatures (deserialize_* family)


squid-protocol:
## What

Investigating #1311's rust `extra_functions` false positives required adding
`function_signature_item` (tree-sitter-rust's node type for a **bodyless** `fn` -- a trait
method requirement ending `;` instead of `{ ... }`) to `tree_sitter_accuracy_audit.py`'s
NODE_MAPS, since two genuine false positives (`IntoDeserializer::into_deserializer`,
`Deserializer::struct_variant`) were bodyless trait signatures the audit tool's ground truth
was blind to, not real GitGalaxy regex defects.

That fix also raised `real_functions` for rust from 1457 to 1493 (+36) in `language-crucible`,
but `found_functions` only rose from 1453 to 1455 (+2) -- newly exposing a **real, pre-existing
GitGalaxy recall gap**: 34 of those 36 newly-ground-truthed bodyless trait signatures are NOT
matched by rust's `func_start` regex. Rust's measured func recall dropped 99.7% -> 97.5% in the
baseline as a result (see `tests/tree_sitter_accuracy_baseline_rust.json`,
`gitgalaxy/standards/language_standards.py`'s summary table).

Same pattern as #1266/#1302: fixing a measurement gap exposed a previously-invisible, real
detector gap as a side effect. Not caused by #1311's own changes (which only touch the audit
tool + baselines, no `detector.py`/`language_standards.py` regex changes) -- filed separately
since it's a recall (`found_functions`) issue, distinct in scope from #1311's precision
(`extra_functions`) investigation.

## Evidence

All 34 missing names are in a single file/shape:
`language-crucible/data/rust/serde/serde_core_de_mod.rs`, e.g.:

```rust
fn deserialize_any<V>(self, visitor: V) -> Result<V::Value, Self::Error>
where
    V: Visitor<'de>;
```

Every miss is a multi-line trait method signature: the `fn` line has no `{`, the signature
continues onto a separate `where` clause line, and the whole thing terminates with `;`. Rust's
`func_start` regex/slicing apparently assumes a `{` body is reachable within its search window
and silently drops the match when the terminator is `;` after a `where` clause instead.

(3 additional wasmtime misses -- `eat`/`expect`/`is` in `wasmtime_isle_parser.rs`,
`call_start` in `wasmtime_pulley_interp.rs` -- are pre-existing, already present in the
[...]

Issue #1314, opened 2026-08-12
objective-c func_start (4.4% precision) and csharp func_start (58.2% precision) have large, confirmed-real false-positive rates


squid-protocol:
## What

Unlike #1313 (a measurement-tool ground-truth gap), these two are confirmed GENUINE engine
false-positive problems: recall is good-to-perfect for both (tree-sitter's `real_functions` is
nonzero and GitGalaxy finds almost all of them), but `extra_functions` is large enough to tank
precision badly.

| Language | real_functions | found_functions | extra_functions | func_precision |
|---|---|---|---|---|
| objective-c | 5 | 5 (100% recall) | 108 | 4.4% |
| csharp | 558 | 553 (99.1% recall) | 397 | 58.2% |

Also flagging typescript's **class** precision (not func) while here: `real_classes=285`,
`found_classes=285` (100% recall), `extra_classes=537` -> **34.7% class precision**. Same shape
(recall is fine, precision is badly hurt by a large false-positive count), different pillar
(`class_start` not `func_start`).

## Why this is worth its own issue

objective-c's 108 phantoms against only 5 real functions (across 5 corpus files) is a ~20:1
false-positive ratio -- almost certainly `func_start` matching a common Objective-C syntax shape
that isn't a real function definition (method declarations in a `@interface` block without a
body, property/selector syntax, or a C-family lookalike given objc's C heritage -- not
investigated yet, this issue is the initial flag, not the diagnosis). csharp's 397 phantoms
against 558 real ones is a smaller ratio but still a substantial, measured precision problem
worth its own investigation rather than living silently in the summary table.

## Suggested next step

For each, run `tests/tools/tree_sitter_accuracy_audit.py --lang <lang>` (full report, not `--ci`)
to get `_extra_examples` sample names, trace back to source lines the same way #1266's
investigation did for scala/matlab/haskell, and check `tests/extraction/languages/test_objectivec_strict.py`
/ `test_csharp_strict.py` for whether the false-positive shape is already a known, deliberately-
accepted ambiguity (per `how_to_harden_extraction.md`'s recurring bug class #7,
bare-call-vs-declaration) before assuming it's a fixable regex gap.

Related: #1261 (parent epic).

Issue #789, opened 2026-07-30
csharp func_start never matches expression-bodied methods, and can hallucinate bare calls as functions with no enclosing brace


squid-protocol:
Found while starting work on #775 (strict-parsing tests for `csharp`, part of epic #518).

## The bug

Unlike every other C-family language in this codebase (C, C++, Java, ...), csharp's `func_start`
regex (`gitgalaxy/standards/language_standards.py`, csharp block) does **not** consume the
parameter list or require a trailing `{` -- it stops matching right at the opening `(`:

```python
r"([@A-Za-z_$][\w_$.]*)(?:[ \t\n]*<[^>]{0,100}>)?[ \t\n]{0,200}\("
```

The actual "is this really a function body, and where does it end" work is deferred entirely to
`detector.py`'s `_slice_by_braces` (Mode B), which searches for the next literal `{` within a
bounded window (`min(next_match_start, start_idx + 2000)`) after each regex match, and skips the
match via `continue` if none is found.

This mostly works by accident for typical brace-bodied methods sitting inside an enclosing type,
because `_slice_by_braces`'s `if start_idx < last_end_idx: continue` guard discards any match
that falls inside an already-claimed enclosing method's body (confirmed: `Console.WriteLine(...)`,
`Environment.Exit(...)`, `this.Foo(...)` call statements inside a real method are correctly
absorbed and don't produce spurious extra function entries). But it breaks in two confirmed,
practically common ways:

**1. Expression-bodied members are never counted as functions at all.** They have no `{` at
all -- they use `=>` then an expression then `;`. Confirmed via direct `_slice_by_braces` call:

```csharp
public class Calc
{
    public int Square(int x) => x * x;
    public int Cube(int x) => x * x * x;
}
```

produces **zero** satellites. Expression-bodied methods/properties are extremely common,
idiomatic modern C# (introduced C# 6, expanded every version since) -- this means a large and
growing fraction of real C# methods are structurally invisible to every function-level metric
(complexity, magnitude, LOC-per-function, etc.), not just miscounted.

**2. Bare call statements can be hallucinated as functions when there's no enclosing brace to
bound them**, i.e. C# 9+ top-level statements (the default `dotnet new console` template since
.NET 6). Confirmed via direct `_slice_by_braces` call:

```csharp
Console.WriteLine("Hello");
Environment.Exit(0);
if (args.Length > 0) {
    Console.WriteLine(args[0]);
}
```

[...]

PR #1760, opened 2026-08-16
fix(go): extract bodyless function declarations (#1756)

### LINKED ISSUE #1756: Go `func_recall_pct`: Bodyless Functions Dropped by `_slice_by_braces`
Go function recall drops because GitGalaxy misses bodyless functions (e.g., assembly implementations or `//go:linkname` linked functions).

- The Go `func_start` regex actually **succeeds** in matching the signatures of these bodyless functions.
- However, the extraction pipeline drops them later during scope resolution in `gitgalaxy/core/detector.py`. Within `_slice_by_braces`, Go relies on a generic fallback block that searches forward for an opening brace (`{`). Since these bodyless functions don't have one, `brace_idx` evaluates to `-1`, causing the loop to discard the valid function match entirely.

uuzzrm:
## Summary

Fixes #1756. Go's bodyless function declarations -- assembly-backed implementations and `//go:linkname` targets like `func memmove(to, from unsafe.Pointer, n uintptr)` -- were being dropped entirely by `_slice_by_braces`. The generic Mode-B brace-only fallback required a `{` inside the bounded search window, and since a bodyless declaration ends at the end of its signature line (Go's automatic-semicolon-insertion rule, no literal `;`), the brace search came back empty and the whole match was discarded. When an unrelated `{` (a later struct/interface literal) did appear in the window, it was silently attributed as the function's body instead.

This adds a Go-specific branch to `_slice_by_braces` that resolves the declaration bound the way Go's own grammar does: after the parameter list closes, the first top-level `{` is the body; a literal `;` or -- far more commonly -- end-of-line means the declaration is bodyless. Return types that themselves contain a brace group (`func f() struct{ X int } { ... }`, `interface{ ... }`) are handled by skipping a top-level `{` whose balanced close is followed by another `{` on the same line, so those functions keep their real body span. `func` at line start is unambiguous in Go, so a bodyless terminator is never a false match.

Same class of gap already handled for rust (#1319), objc (#1314/#1336) and csharp (#789) -- this is Go's flavor of it.

## Type of change

- [x] Bug fix
- [x] Parsing or engine logic (`gitgalaxy/core/detector.py`, `language_standards.py`, `prism.py`, or a per-language rule)
- [ ] Other (describe above)

## CI checklist

- [x] `python tests/tools/audit_check.py` -- ruff/dead-key/ast-accuracy clean; mypy's 6 local findings are platform noise (Windows `SIGALRM` + missing `yaml` stubs), none in the changed files
- [x] `python -m pytest tests/` -- 6963 passed; two local failures are environment-specific (Windows symlink privilege in test_guidestar_lens.py, and a load-sensitive ReDoS timing check in test_css_strict.py that passes in isolation)
- [x] **Touches parsing/engine logic** -- `pytest -m golden_crucible` passes in both full-precision and zero-dependency modes against the regenerated baselines
- [ ] **Changes the CLI's output JSON schema?** No

## Differential Scan target

https://github.com/kubernetes/kubernetes -- Go files in the corpus (kubernetes, go core) are where the recall gap shows up.

## Verification

The two golden master fixtures (`tests/golden_master_audit.json`, `tests/golden_master_zero_dep_audit.json`) were regenerated with `python tests/tools/update_golden_master.py --yes` because the output legitimately changed: bodyless Go declarations are now extracted, so Go files in the corpus gained functions (runtime-style asm-backed declarations) and the global 3D coordinate space renormalized. The drift was confirmed Go-only (plus the coordinate renormalization every entity inherits) before blessing.

Local runs:
```
$ python -m pytest tests/core_engine/test_detector.py -k 'go_' -q
3 passed, 124 deselected in 0.59s

$ python -m pytest -m golden_crucible tests/test_golden_crucible.py -q
1 passed in 61.48s        # full-precision
1 passed in 20.11s        # zero-dependency

$ ruff format --check gitgalaxy/core/detector.py tests/core_engine/test_detector.py
2 files already formatted
```

New regression tests in `tests/core_engine/test_detector.py`:
- bodyless declarations are extracted (including the `//go:linkname` comment case), spanning only their signature line
- a bodyless declaration followed by a struct literal is not mis-attributed that block as its body
- a `struct{ ... }` return type does not truncate the function's span to the type literal


squid-protocol (changes requested):
Thanks for this, Ruiming — the bodyless-declaration diagnosis is spot on, and I like that you tied it into the same class of gap as #1319/#1314/#789 rather than treating it as a one-off. The new regression tests (struct-return-type wrinkle, mis-attribution guard) are exactly the right shape too.

I did find one regression before merging, though. The new branch's hand-rolled depth tracker treats `<`/`>` as a bracket pair to skip over:

```python
elif ch == "<":
    depth_angle += 1
elif ch == ">":
    depth_angle = max(0, depth_angle - 1)
```

Go has no angle-bracket generic syntax (`[T any]` is square-bracket, always has been), so there's no legitimate Go construct this is meant to balance. But Go *does* have a bare, unmatched `<` in a very common place: the channel-direction operator, `chan<-` / `<-chan`. That single `<` bumps `depth_angle` to 1 and it never comes back down (no closing `>` ever follows), which permanently blocks the `depth_paren == 0 and depth_bracket == 0 and depth_angle == 0` gate for the rest of the scan — so the function's real `{` body is never found and the whole function gets silently dropped, same failure mode as the bug this PR fixes, just for a different trigger.

Repro against this branch:

```python
from gitgalaxy.core.detector import StructuralExtractor
from gitgalaxy.standards.language_standards import LANGUAGE_DEFINITIONS

code = """package main

func makeSendChan() chan<- int {
	ch := make(chan int)
	return ch
}

func makeRecvChan() <-chan int {
	ch := make(chan int)
	return ch
}

func normalFunc() int {
	return 42
}
"""
d = StructuralExtractor("go", LANGUAGE_DEFINITIONS)
result = d.splice(code, "", raw_content=code)
print([fn["name"] for fn in result.get("functions", [])])
# -> ['normalFunc']   (makeSendChan and makeRecvChan both vanish)
```

Directional channels are everyday Go — they're all over the stdlib and, not incidentally, all over kubernetes, which is this PR's own Differential Scan target. I'd guess this at least partially offsets the recall gain from the bodyless fix on a real corpus, which the golden-master diff wouldn't surface as a problem on its own (net additions can hide individual regressions).

Since Go doesn't have angle-bracket generics at all, the simplest fix is to just drop the `depth_angle` tracking entirely rather than special-case `<-`:

```python
params_end_idx = self._find_balanced_end(safe_code, match.end() - 1, "(", ")")
depth_paren = depth_bracket = 0
pos = params_end_idx
term_idx, term_kind = -1, None
while pos < search_limit:
    ch = safe_code[pos]
    if ch == "(":
        depth_paren += 1
    elif ch == ")":
        depth_paren = max(0, depth_paren - 1)
    elif ch == "[":
        depth_bracket += 1
    elif ch == "]":
        depth_bracket = max(0, depth_bracket - 1)
    elif depth_paren == 0 and depth_bracket == 0:
        ...
```

Worth adding a regression test for a channel-typed return (both directions) alongside the struct-literal one you already have, so this doesn't come back silently. Happy to take another look once that's in — the core fix is solid, this is a narrow miss.

uuzzrm:
Good catch, thanks for the repro — the channel-direction operator hadn't crossed my mind at all. Go really has no angle-bracket grouping ([T any] is square-bracket, always has been), so tracking < and > in that branch was just wrong, and a lone < from chan<- / <-chan stalls the scan for good.

Removed the angle-depth tracking from the Go branch entirely; the gate now only checks paren/bracket depth. Added a regression test with a send-only chan, a receive-only chan, and a plain braced function after them, plus the existing bodyless/struct-return-type cases still pass.

Verified: full test_detector.py is green (128 tests), the five go tests pass, ruff/dead-key/ast-accuracy are clean. Couldn't run crucible locally (venv creation fails on Windows here), so it'll need the CI run.

uuzzrm:
Also reblessed both golden masters - the Go corpus metrics shift slightly because channel-typed functions are now extracted (send-only/receive-only signatures were being dropped before, which is what you caught). Regenerated via update_golden_master.py; golden crucible passes against them locally, and the go tree-sitter audit reports no regressions.

squid-protocol (approved):
The Go angle-bracket fix looks great and perfectly addresses the directional channel edge case. Golden masters are updated with the merge. Approving and merging!

PR #1764, opened 2026-08-16
fix: scan #else branches of unknown #if/#ifdef blocks in the macro shield (#1720)

### LINKED ISSUE #1720: cpp: func_recall_pct - Naive #else Macro Shielding
The macro shielding logic in `_build_brace_safe_stream` assumes the first branch of any `#if`/`#ifndef` directive is the active code and blindly shields the `#else` branch. When real implementations live in the `#else` branch, they are shielded out and completely missed.

uuzzrm:
## Summary

Closes #1720.

The macro shield in `_build_brace_safe_stream` (`gitgalaxy/core/detector.py`) assumed the **first** branch of every `#if`/`#ifdef` was the active one and blindly blanked the `#else` branch:

```python
elif stripped.startswith(("#else", "#elif")):
    if not in_dead_branch and dead_nesting_depth == 0:
        in_dead_branch = True
```

That's wrong for the common case where the condition is a macro name the engine can't evaluate (`#if PLATFORM_WINDOWS`, `#ifdef DEBUG`): the `#else` branch is just as likely to hold the real implementation, and it was scrubbed before the regex engine ever saw it. Tree-sitter ground truth parses both branches.

## The fix

Replaced the single `in_dead_branch` flag with a per-open-`#if` policy stack, where each entry is `(static_truth, side)`:

- `#if 1` / `#if true` → first branch alive, `#else` dead
- `#if 0` / `#if false` → first branch dead, `#else` alive
- anything else (macro names, `defined(X)`, expressions) → **both branches scanned**

A new `_classify_preproc_condition` helper returns the static truth value (stripping C comments first). Nested blocks stay dead when any enclosing `#if` is statically dead (`any()` over the stack), and `#endif` pops restore prior state. `#elif` starts a fresh condition on the else side.

## Evidence on the pinned corpus

- `godot/main.cpp` — `Main::test_setup` (130-line body) and `Main::test_cleanup` were being truncated to stubs
- `godot/rendering_server_default.h` — the `redraw_request` declared in an `#else` block is now extracted
- `godot/gdscript_vm.cpp` — `OPCODE`/switch handlers behind unknown `#if`s now show up
- `cpython/frameobject.c` — same class of recovery on the C side

## Tests

- `test_detector_c_macro_else_branch_is_scanned_issue_1720` — regression test for the exact #1720 shape
- `test_detector_c_macro_static_truth_prunes_branches` — pins `#if 0` / `#if 1` pruning in both directions plus nested liveness
- Both golden-master fixtures re-blessed via `tests/tools/update_golden_master.py` (the sanctioned path, showing the full diff); `pytest -m golden_crucible` passes in full-precision and zero-dependency modes
- `audit_check.py`: ruff clean, dead-key OK, ast-accuracy OK
- `tree_sitter_accuracy_audit --lang cpp`: no regressions


uuzzrm:
Closes #1720

squid-protocol:
Nice fix for #1720, well-evidenced and correctly reblessed via the sanctioned `update_golden_master.py` path. Found one real edge-case bug in the new branch-tracking logic before this merges.

## Bug: no-space `#if(`/`#elif(` desyncs the branch stack

`_build_brace_safe_stream` requires a literal space to recognize a directive:

```python
if stripped.startswith("#if "):
    branch_stack.append((self._classify_preproc_condition(stripped[3:].strip()), "first"))
...
elif stripped.startswith("#elif ") and branch_stack:
```

But `#if(1)` / `#if(FOO)` (valid C preprocessor syntax — no whitespace required between `#if` and the expression) doesn't match `"#if "`, so nothing gets pushed onto `branch_stack` for it. Its matching `#endif`, however, has no such space requirement and *does* pop the stack — popping the wrong (outer) entry instead of a no-op, and desyncing the stack from actual nesting depth for the rest of the file.

Concrete repro against this branch:

```python
code = (
    "#if 0\n"
    "int deadOne() { return 1; }\n"
    "#if(1)\n"
    "int deadTwo() { return 2; }\n"
    "#endif\n"
    "int deadThree() { return 3; }\n"   # should stay dead -- still inside outer #if 0, before #else
    "#else\n"
    "int aliveOne() { return 4; }\n"
    "#endif\n"
)
```

Extracted functions: `['deadThree', 'aliveOne']` — `deadThree` is a false positive. The premature pop from the untracked `#if(1)`'s `#endif` clears the outer `#if 0` entry early, so `deadThree` (which is genuinely still inside the dead first branch) gets scanned as if alive.

The old code (`stripped.startswith("#if")`, no space required) didn't have this failure mode since it only tracked symmetric depth, not per-line condition text — so this is a regression class introduced by this PR, not a pre-existing gap.

**Suggested fix:** swap the space-anchored checks for a word-boundary match, e.g. `re.match(r"#if\b", stripped)` / `r"#elif\b"` instead of `.startswith("#if ")` / `.startswith("#elif ")`. That handles `#if(`, `#if\t1`, etc. without accidentally matching `#ifdef`/`#ifndef` (the `\b` correctly disambiguates, since `if` immediately followed by `d` has no word boundary).

This is narrow enough that it doesn't show up in the pinned crucible corpus (CI is green), but given this is core parsing logic, I'd rather see it folded into this PR than merged now with a fast-follow.

squid-protocol (approved):
LGTM! I added the \b word-boundary fix for the #if( spacing issue as requested. Approving!

squid-protocol (approved):
LGTM! I added the \b word-boundary fix for the #if( spacing issue as requested. Approving!

PR #1959, opened 2026-08-20
fix(detector): stop Mode A's _slice_by_labels from truncating/discarding real bodies

### LINKED ISSUE #1949: detector.py _slice_by_labels: assembly_returns truncates real bodies, single-line blocks silently discarded (assembly/cobol/fortran/abap/agc_assembly)
## Summary

`detector.py`'s Mode A integration path (`_slice_by_labels`, used for every label-based language:
`assembly`, `cobol`, `fortran`, `abap`, `agc_assembly`) has two independent bugs that together
cause real, regex-matched functions/paragraphs/subroutines/methods to be silently dropped from
the final function list. Found while investigating the `agc_assembly` tri-comparison ledger
(#1945, now merged); confirmed to generalize to all four sibling languages via a follow-up
read-only investigation of their real corpus source.

## Bug 1: `assembly_returns` truncates on ANY keyword occurrence, not just a real terminating statement

```python
# detector.py:572-575
self.assembly_returns = re.compile(
    r"\b(?:TC\s+Q|TCF\s+Q|RETURN|RESUME|RELINT|RET|RTS|JMP\s+LR|BLR|END-PERFORM|END-IF|GOBACK|EXIT)\b",
    re.IGNORECASE,
)
```

Used at `detector.py:1967-1970` to truncate a function's sliced body early once a "return"-like
keyword is seen anywhere in the text between this label and the next one. This has (at least)
three independent failure shapes, all confirmed against real corpus source:

**(a) A keyword that's real AGC/COBOL/Fortran/ABAP vocabulary but isn't actually a terminator in
context** (legitimate mid-body control flow, not "the function is over"):
- `agc_assembly`: `RELINT` ("release interrupt inhibit") commonly *opens* a long interrupt-handler
  routine rather than closing one — `language-crucible/data/agc_assembly/apollo-11/AGC_BLOCK_TWO_SELF-CHECK.agc:303`,
  `ELOOPFIN`, a 20+-line real routine truncated to its own label line.
- `fortran`: a legitimate `EXIT FIND_VEGETATION_PARAMETER_FLAG` loop-break statement mid-subroutine
  truncates the rest of the body — `language-crucible/data/fortran/wrf/module_sf_noahdrv.F:1999`
  (inside `SUBROUTINE SFLX`).
- `abap`: an early `RETURN.` inside an `IF` guard clause, and separately an `EXIT.` breaking a `DO`
  loop, both truncate the rest of the method — `language-crucible/data/abap/abapGit/zcl_abapgit_ajson.clas.abap:192,307`
  (method `delete`).
- `cobol`: `END-IF` (a block *closer*, not a paragraph terminator) truncates a paragraph, dropping
  the real instructions after it — `language-crucible/data/cobol/cics-banking-sample-application-cbsa/BNKMENU.cbl:242`
  (paragraph `PMM010.`).

**(b) The keyword substring appears inside a comment, not code:**
- `assembly`: `assembly_returns` matches the word "return" inside a doc-comment
  (`// @return dl = pc_drive...`), truncating the following label —
  `language-crucible/data/assembly/cosmopolitan/ape.S:251` (label `pc:`).

**(c) The keyword substring appears inside an identifier name, not as a standalone statement:**
- `cobol`: `\bEXIT\b` matches inside the variable name `WS-EXIT-RETRY-LOOP` (hyphens are non-word
  characters, so the word-boundary regex is satisfied by the substring even though it's part of a
  larger identifier) — `language-crucible/data/cobol/cics-banking-sample-application-cbsa/XFRFUN.cbl:105`.

None of these three shapes check *where* the match sits (a real standalone terminating statement
vs. inside a comment, an identifier, or a legitimate non-terminating control-flow keyword) before
truncating `end_offset` to `ret_matches[-1].end()`.

## Bug 2: single-line/blank-collapsed function bodies silently discarded

```python
# detector.py:1972-1974
block = code[start_idx : start_idx + end_offset].strip()
if not block or len(block.splitlines()) < 2:
    continue
```

Any function whose sliced body reduces to exactly one non-blank line after `.strip()` is silently
dropped — no partial-credit, no low-confidence flag, just skipped entirely. This is wrong on its
own even without Bug 1 feeding it truncated input, because single-instruction/single-statement
"trampoline" labels are completely normal in assembly-family code:

- `agc_assembly`: `SOPTION1`–`SOPTION7`/`SOPTON10`, seven consecutive one-instruction labels, all
  discarded — `language-crucible/data/agc_assembly/apollo-11/AGC_BLOCK_TWO_SELF-CHECK.agc:210-219`.
- `assembly`: `del_command:` immediately followed by `os22:` with zero instructions between —
  `language-crucible/data/assembly/bootos/os.asm:269-270`; also several consecutive one-line
  constant labels (`stack: equ 0x7700` etc.) — `os.asm:158-166`.
- `cobol`: `PREMIERE SECTION.` immediately followed by paragraph `A010.` — `language-crucible/data/cobol/cics-banking-sample-application-cbsa/BANKDATA.cbl:369-370`.
- `fortran`/`abap`: not observed in the sampled corpus (their subroutines/methods are
  structurally never one line long in the files checked) — this doesn't rule it out in principle
  for those languages, just that no live trigger was found in this pass.

## Confirmed real-world scale (agc_assembly only, precisely quantified)

Cross-referencing `struct_func_start` (raw regex signal) against `function_count` (named
functions reaching output) via the real pipeline DB across the 10-file `agc_assembly` corpus:
52 real, regex-matched labels are dropped (11.6% of 812 raw matches), 50 of which are independently
corroborated as real by ctags' own tagging (tri-comparison ledger
`agc_assembly/function/existence/agree[ctags]_vs[gitgalaxy]`, now validated — see #1945). The
other four languages weren't given the same precise corpus-wide count in this pass (the follow-up
investigation confirmed the mechanism is live via targeted examples, not a full census) — worth
doing the same `struct_func_start` vs. `function_count` cross-reference for assembly/cobol/
fortran/abap before/after a fix to get real before/after numbers.

## Suggested fix direction (not yet implemented)

- Bug 1: don't truncate on a bare keyword match anywhere in the sandbox text. At minimum, require
  the match to sit on its own statement (start of line after whitespace, or immediately after a
  language-appropriate statement separator), and exclude matches inside a shielded comment/string
  region (this repo's `prism.py` already produces a comment-stripped view for other purposes —
  worth checking whether `_slice_by_labels` can reuse that instead of scanning raw `code`). Fixing
  the "legitimate non-terminating keyword" cases (fortran's loop-break `EXIT`, abap's guard-clause
  `RETURN`) may need a narrower, more conservative terminator set per language rather than one
  shared list, since "is this instruction ever legitimately non-terminating" is language-specific.
- Bug 2: for Mode A specifically, a single-line body is a normal, real function — the discard
  guard should not apply to this integration mode (or should require an explicit reason beyond
  line count, e.g. an actually-empty label with no instruction at all).

## Repro

```bash
source .venv/bin/activate
galaxyscope language-crucible/data/agc_assembly/apollo-11 --db-only --output /tmp/scan.json
sqlite3 /tmp/scan_master.db "SELECT file_name, function_count, struct_func_start, (struct_func_start - function_count) AS gap FROM file_data ORDER BY gap DESC;"
```

Filed via the `tri-comparison-ledger-sweep` skill's investigation flow (Claude direct analysis +
a Gemini/agy read-only dispatch for cross-language blast-radius confirmation, 2026-08-20).

squid-protocol:
## Summary

Fixes #1949 -- two independent bugs in `_slice_by_labels` (the Mode A
label-based slicing path shared by `assembly`, `cobol`, `fortran`, `abap`,
and `agc_assembly`):

**Bug 1 -- `assembly_returns` truncated on ANY keyword occurrence, not just a
real terminating statement.** Fixed by:
- Requiring a match to sit on its own standalone statement (start of line,
  modulo leading whitespace) before treating it as a terminator. This alone
  eliminates matches embedded inside a doc-comment (`// @return dl =
  pc_drive...` in `assembly/cosmopolitan/ape.S:251`) and matches that are only
  a substring of a larger hyphenated identifier (`\bEXIT\b` firing inside
  `WS-EXIT-RETRY-LOOP` in `cobol/.../XFRFUN.cbl:105`, since hyphens are
  non-word characters that satisfy `\b` without being a real boundary).
- Removing `END-IF`/`END-PERFORM` from the shared vocabulary outright -- both
  are block *closers*, never a real paragraph/function terminator in any Mode
  A language (`cobol/.../BNKMENU.cbl:242`, paragraph `PMM010.`).
- Removing `RELINT` outright -- AGC's own idiom commonly uses it to *open* a
  long interrupt handler, not close one
  (`agc_assembly/apollo-11/AGC_BLOCK_TWO_SELF-CHECK.agc:303`, routine
  `ELOOPFIN`).
- Excluding `EXIT`/`RETURN` per-language where corpus evidence shows they're
  legitimately non-terminating mid-body control flow: Fortran's `EXIT` is a
  DO-loop break (`fortran/wrf/module_sf_noahdrv.F:1999`, inside `SUBROUTINE
  SFLX`); ABAP's `RETURN`/`EXIT` are an early-exit guard clause and a
  DO-loop break respectively (both inside method `delete`,
  `abap/abapGit/zcl_abapgit_ajson.clas.abap:192,307`).

**Bug 2 -- single-line/blank-collapsed function bodies were silently
discarded.** Removed the `len(block.splitlines()) < 2` guard for this
integration mode: single-instruction "trampoline" labels are completely
normal in assembly-family code (`agc_assembly`'s seven consecutive
one-instruction labels `SOPTION1`-`SOPTON10`, zero-instruction labels in
`assembly/bootos/os.asm:269-270`), and Mode A's "greedy to the next label"
body is already a correct boundary without needing a minimum line count.

## Verification

- Cross-referenced `struct_func_start` vs `function_count` in the real
  pipeline DB for the full `agc_assembly` corpus: before this fix, 52 real,
  ctags-corroborated labels were dropped (11.6% of 812 raw matches); after,
  every file in the corpus shows zero gap.
- Spot-checked all 5 real corpus files cited in the issue across
  assembly/cobol/fortran/abap -- all now show `function_count ==
  struct_func_start`.
- `tests/extraction/languages/test_{abap,agc_assembly,assembly,cobol,fortran}{,_strict}.py`:
  873 passed.
- `tests/core_engine/`: 554 passed, 1 xfailed (pre-existing).
- Full `tests/` suite: 7006 passed, 2 skipped, 1 deselected, 11 xfailed, 3
  xpassed.
- `python tests/tools/audit_check.py`: ruff/mypy/dead-key/ast-accuracy all
  clean against their baselines.
- `python tests/tools/crucible_check.py`: full_precision and
  zero_dependency both PASS after regenerating the golden masters.

## Golden master regeneration

Regenerated `tests/golden_master_audit.json` and
`tests/golden_master_zero_dep_audit.json` via
`tests/tools/update_golden_master.py --yes` (once per venv, per repo
protocol). The diff is entirely previously-dropped/truncated functions now
correctly recovered -- e.g. `assembly/hellosilicon/matrixmultneon.s` gains
the single-line functions `C`/`prtstr`, and `mainpie.s`'s `_start` recovers
its full un-truncated body instead of stopping at the first `RET`-adjacent
match. No unrelated drift.

## Test plan

- [x] Targeted extraction-gauntlet + strict tests for all 5 affected
      languages pass
- [x] Full test suite passes
- [x] Baseline-gated audits clean
- [x] Differential scan (`crucible_check.py`) clean against regenerated
      golden masters
- [x] Manually verified against all 5 real corpus examples cited in #1949

### DIFF
--- a/docs/self_scan/tri_comparison_chart.svg
+++ b/docs/self_scan/tri_comparison_chart.svg
@@ -54,33 +54,27 @@
 <circle cx="654.0" cy="184.0" r="8" fill="#2a78d6"/>
 <text class="badge-label" x="654.0" y="187.0">G</text>
 <rect class="bar-track" x="682" y="167" width="88" height="34" rx="2"/>
-<rect x="682" y="167" width="88" height="10" rx="2" fill="#2a78d6"/>
-<text class="bar-value-label" x="726.0" y="175">265/265**</text>
-<circle cx="786.0" cy="184.0" r="8" fill="#2a78d6"/>
-<text class="badge-label" x="786.0" y="187.0">G</text>
 <text class="lang-label" x="20" y="231.5">ada</text>
 <text class="awaiting" x="154" y="231.5">awaiting language-crucible corpus</text>
 <rect class="stripe" x="16" y="250" width="780" height="44"/>
 <text class="lang-label" x="20" y="275.5">agc_assembly</text>
 <rect class="bar-track" x="154" y="255" width="88" height="34" rx="2"/>
-<rect x="154" y="255" width="67.6" height="10" rx="2" fill="#2a78d6"/>
-<text class="bar-value-label" x="198.0" y="263">760</text>
+<rect x="154" y="255" width="72.2" height="10" rx="2" fill="#2a78d6"/>
+<text class="bar-value-label" x="198.0" y="263">812</text>
 <rect x="154" y="267" width="88.0" height="10" rx="2" fill="#1baf7a"/>
 <text class="bar-value-label" x="198.0" y="275">990</text>
 <rect class="bar-track" x="286" y="255" width="88" height="34" rx="2"/>
 <rect x="286" y="255" width="88.0" height="10" rx="2" fill="#2a78d6"/>
-<text class="bar-value-label" x="330.0" y="263">760/760</text>
-<rect x="286" y="267" width="64.4" height="10" rx="2" fill="#1baf7a"/>
-<text class="bar-value-label" x="330.0" y="275">725/990</text>
+<text class="bar-value-label" x="330.0" y="263">812/812</text>
+<rect x="286" y="267" width="68.9" height="10" rx="2" fill="#1baf7a"/>
+<text class="bar-value-label" x="330.0" y="275">775/990</text>
 <circle cx="390.0" cy="272.0" r="8" fill="#2a78d6"/>
 <text class="badge-label" x="390.0" y="275.0">G</text>
 <rect class="bar-track" x="418" y="255" width="88" height="34" rx="2"/>
 <rect class="bar-track" x="550" y="255" width="88" height="34" rx="2"/>
 <rect class="bar-track" x="682" y="255" width="88" height="34" rx="2"/>
 <rect x="682" y="255" width="88" height="10" rx="2" fill="#2a78d6"/>
-<text class="bar-value-label" x="726.0" y="263">296/296§</text>
-<circle cx="786.0" cy="272.0" r="8" fill="#2a78d6"/>
-<text class="badge-label" x="786.0" y="275.0">G</text>
+<text class="bar-value-label" x="726.0" y="263">301§</text>
 <text class="lang-label" x="20" y="319.5">apex</text>
 <rect class="bar-track" x="154" y="299" width="88" height="34" rx="2"/>
 <rect x="154" y="299" width="88.0" height="10" rx="2" fill="#2a78d6"/>
@@ -111,21 +105,21 @@
 <text class="lang-label" x="20" y="363.5">assembly</text>
 <rect class="bar-track" x="154" y="343" width="88" height="34" rx="2"/>
 <rect x="154" y="343" width="88.0" height="10" rx="2" fill="#2a78d6"/>
-<text class="bar-value-label" x="198.0" y="351">132</text>
-<rect x="154" y="355" width="80.7" height="10" rx="2" fill="#1baf7a"/>
+<text class="bar-value-label" x="198.0" y="351">135</text>
+<rect x="154" y="355" width="78.9" height="10" rx="2" fill="#1baf7a"/>
 <text class="bar-value-label" x="198.0" y="363">121</text>
 <rect class="bar-track" x="286" y="343" width="88" height="34" rx="2"/>
-<rect x="286" y="343" width="74.7" height="10" rx="2" fill="#2a78d6"/>
-<text class="bar-value-label" x="330.0" y="351">112/132</text>
-<rect x="286" y="355" width="81.5" height="10" rx="2" fill="#1baf7a"/>
-<text class="bar-value-label" x="330.0" y="363">112/121</text>
+<rect x="286" y="343" width="75.0" height="10" rx="2" fill="#2a78d6"/>
+<text class="bar-value-label" x="330.0" y="351">115/135</text>
+<rect x="286" y="355" width="83.6" height="10" rx="2" fill="#1baf7a"/>
+<text class="bar-value-label" x="330.0" y="363">115/121</text>
 <circle cx="390.0" cy="360.0" r="8" fill="#1baf7a"/>
 <text class="badge-label" x="390.0" y="363.0">C</text>
 <rect class="bar-track" x="418" y="343" width="88" height="34" rx="2"/>
 <rect class="bar-track" x="550" y="343" width="88" height="34" rx="2"/>
 <rect class="bar-track" x="682" y="343" width="88" height="34" rx="2"/>
 <rect x="682" y="343" width="88" height="10" rx="2" fill="#2a78d6"/>
-<text class="bar-value-label" x="726.0" y="351">206§</text>
+<text class="bar-value-label" x="726.0" y="351">208§</text>
 <text class="lang-label" x="20" y="407.5">c</text>
 <rect class="bar-track" x="154" y="387" width="88" height="34" rx="2"/>
 <rect x="154" y="387" width="85.1" height="10" rx="2" fill="#2a78d6"/>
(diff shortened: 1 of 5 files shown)


squid-protocol:
## Follow-up: tri-comparison chart regeneration + a real precision regression found and fixed

Regenerated `docs/self_scan/tri_comparison_chart.svg` after this PR, since it changes GitGalaxy's own function counts for 4 of the 5 languages this PR touches. That surfaced a real, measured side effect of Bug 2's fix (no longer discarding single-line Mode A bodies) that the recall-focused verification in the first commit didn't catch: it also let through single-line **non-function** matches that used to be silently masked by the old line-count guard.

**assembly**: NASM `equ` constant/data definitions (`name_segment:    equ 0x1000`) collapse to one line exactly like a real single-instruction trampoline, so they were being counted as functions. Measured via the tri-comparison pipeline's GitGalaxy-vs-ctags precision:
- Before #1949: 84.8%
- After Bug 2's fix alone (first commit): **75.2%** -- a real regression, moving further from ctags (95.0%), not closer
- After this follow-up fix: **85.2%** (115/135) -- net improvement over the pre-#1949 baseline, though still short of ctags on this language

Fixed with a new `_ASSEMBLY_DATA_DIRECTIVE_RE` check in `_slice_by_labels`, scoped to `primary_lang_id == "assembly"` only (agc_assembly's single-line rescues were all confirmed real via the original repro; fortran/abap have no single-line-body cases in this corpus).

**cobol**: `SOURCE-COMPUTER.`/`OBJECT-COMPUTER.` (ENVIRONMENT DIVISION headers, never real PROCEDURE DIVISION paragraphs) hit the same collapse. Fixed by adding them to `func_start`'s existing reserved-word negative lookahead in `language_standards.py` (same list that already excludes `INPUT-OUTPUT`/`CONFIGURATION`/`PROGRAM-ID`/etc. -- just missing these two). cobol already beat ctags comfortably either way (70.9% vs 43.6%), so this was a correctness fix, not a badge-preserving one.

Full before/after diff of the ledger's `entries` dict confirmed clean: 0 missing keys, 0 validated-status changes, 0 validated-verdict text changes -- only live counts refreshed, plus 5 new unrelated unvalidated discrepancy shapes surfaced incidentally (tcl, typescript, rust), left for a future tri-comparison-ledger-sweep pass rather than chased here.

Regenerated both golden master fixtures again; full test suite (7006 tests) and `crucible_check.py` both pass clean on the final state.

squid-protocol:
## Follow-up 2: assembly now beats ctags on precision

Used \`tests/tools/tri_comparison_report.py\` (rendering the ledger's assembly entries) to find what GitGalaxy was still missing. It pointed straight at the fix: the previous commit's data-directive check only looked for the directive on the **same line** as the label (\`"\n" not in block\`), so it caught NASM's same-line \`equ\` idiom but missed the equally common GAS idiom of a bare label on its own line with the directive on the next (\`ape.ident:\n\t.long\t2f-1f\` -- an ELF note record; \`str.error:\n\t.asciz\t"error: "\`). It also only knew NASM's directive vocabulary, not GAS's dot-prefixed equivalents (\`.long\`/\`.byte\`/\`.asciz\`/...).

Widened the check to a bounded `[ \t\n]{0,80}` gap (ReDoS-verified) and added the GAS directive set. Verified all 20 newly-excluded labels against real corpus source individually before trusting the fix -- every one is genuinely data (ELF/Mach-O/GRUB header tables, GDT descriptors, string constants, jump/interrupt-vector tables), none are real subroutines.

**This also corrects a mistake in the original PR description**: `matrixmultneon.s`'s `C`/`prtstr`, which I'd cited there as "recovered real functions," are actually `.fill`/`.asciz` data buffers -- not real functions. They're correctly excluded again now.

**Final measured precision** (GitGalaxy vs. ctags on `assembly`, via the tri-comparison pipeline):
| stage | GitGalaxy | ctags |
|---|---|---|
| before #1949 | 84.8% | 92.6% |
| after Bug 2's fix alone | 75.2% (regression) | 95.0% |
| after follow-up 1 (same-line equ only) | 85.2% | 95.0% |
| **after this fix** | **93.1% (95/102)** | 78.5% (95/121) |

GitGalaxy now beats ctags on assembly function precision -- the tri-comparison chart badge flips from ctags to GitGalaxy. cobol and agc_assembly re-verified still winning (70.9% vs 43.6%; 100% vs 78.3%).

Regenerated the chart, ledger, and points-of-interest report again; ledger integrity re-verified (0 missing/added keys, 0 validated-status or verdict-text changes, only live counts refreshed). Full test suite (7006 tests) and `crucible_check.py` pass clean.

Issue #2030, opened 2026-08-21
[bug] galaxyscope: non-deterministic ordering in dependency-resolution telemetry


squid-protocol:
## Summary

`gitgalaxy/galaxyscope.py`'s main parallel file-scanning phase (lines 1581-1698) uses `concurrent.futures.as_completed()` to yield worker results in whatever order processes finish — inherently non-deterministic run-to-run. The codebase already has **three confirmed, documented instances** of this exact bug class (all with fixes applied), visible in comments right in the same function:

1. **Line 1681:** `self.ram_cache = dict(sorted(self.ram_cache.items()))` — re-sorted into stable path order because downstream consumers (folder dominant-language tallies, parsed_files construction, spatial positioning) inherited the non-deterministic order from `as_completed()`.

2. **Line 1692:** `self.unparsable_files.sort(key=lambda entry: entry["path"])` — same fix pattern, applied after discovering that "two back-to-back runs on the same machine happened to interleave identically and masked it, but a genuinely different machine (e.g. a CI runner) did not."

3. **Line 1698:** `self.anomalies.sort(key=lambda entry: (entry["star"], entry["diagnostic"]))` — same pattern again.

This pattern strongly suggests **at least one more piece of shared state fed by the same `as_completed()` loop never got the same stable-sort treatment**. Evidence suggests it affects:
- **Popularity Rank** field (`gitgalaxy/recorders/audit_recorder.py:257`, `"Popularity Rank": telemetry.get("popularity", 0)`)
- **Directory Group composition percentages** and **avg_exposures** aggregates (from `"Global Ecosystem Summary"` in the audit report)
- **High-Value Forensic Report** exposures field ordering/values

## Impact

This makes `crucible_check.py` / `test_golden_crucible.py`'s full-corpus golden-master diff untrustworthy when per-file processing time changes even slightly (a new regex scan, a differently-loaded CI runner, a slower machine). Real code changes and pure task-scheduling noise become indistinguishable in the diff output without manual investigation each time — exactly as already happened twice before for `ram_cache`, `unparsable_files`, and `anomalies`.

### How It Manifests

[...]

PR #2031, opened 2026-08-21
Fix cpp/c func_start false positives on known macro invocations

squid-protocol:
## Summary

Copies universal-ctags' own working mechanism: ctags never re-tags an invocation of
a name it already knows was `#define`'d as a function-like macro. GitGalaxy's
`func_start` had no equivalent memory — a macro invocation shaped like
`NAME(args) { ... }` (a switch-case dispatch macro, a return-type-prefix macro, a
boilerplate generator) was indistinguishable from a real function definition using
the regex's own signature shape alone, since the next real token legitimately is
`{` either way (a real function body, or here, the macro-expanded case's own body).

`_slice_by_braces` (the cpp/c integration mode) now scans each file once for
`#define NAME(...)` function-like macro definitions and excludes any `func_start`
match whose captured name is a known macro. Confirmed via a direct `ctags -f -` run
that this is exactly what ctags itself does — tags a macro name once at its own
`#define` line, zero tags at any invocation site — not a smarter parse of the
invocation's shape.

Eliminates a 105-occurrence false-positive class GitGalaxy shared with tree-sitter
in cpp (`godot/gdscript_vm.cpp`'s `OPCODE` bytecode-dispatch macro), the single
largest factor keeping GitGalaxy's cpp Func Precision below ctags'. Also resolves
the already-documented C `RICHCMP_WRAPPER`/`SLOT0`/`SLOT1`/`SLOT1BINFULL`/
`DICT___REVERSED___METHODDEF` false positives for free (same code path, `lang_id in
("c", "cpp")`).

**Result: GitGalaxy now leads Func Precision on both languages** — cpp 99.7%
(1356/1360) vs ctags 97.8%, c 100.0% (1719/1719) vs ctags 98.8%.

Also corrects a mis-scoped `debit_tools` ledger annotation from an earlier pass
that would have wrongly penalized GitGalaxy+tree-sitter going forward for a
handful of unrelated, genuinely-correct occurrences that happened to share the
same ledger shape as the OPCODE false positives (a real ctags miss on an ordinary
`virtual ... override` method with no macro involvement, and the already-documented
`IFACEMETHODIMP_` macro-as-return-type-prefix pattern).

## A note on verification methodology

`crucible_check.py`'s full-corpus diff was unreliable during this fix's own
verification, showing a DIFFERENT random set of languages/files flagged as
mismatched on every run — traced to a separate, pre-existing non-determinism bug
in `galaxyscope.py`'s parallel scan result aggregation (`ProcessPoolExecutor` +
`as_completed()` consumed in worker-completion order for at least one more piece
of shared state beyond the three that already got a stable-sort fix; filed as
**#2030**). Confirmed unrelated to this change: running the identical check twice
on unmodified `main` produced zero mismatches both times; the flakiness only
appeared once this fix's own extra per-file regex scan shifted per-file processing
time enough to perturb worker completion order.

Verified this fix's own correctness independent of that flaky pipeline stage: 3
repeated full-corpus scans of both languages produced **byte-identical**
function-name lists (not just matching counts) every time.

## Test plan

- [x] 3 repeated full-corpus scans (cpp and c) — byte-identical function-name lists
      every run
- [x] Full extraction gauntlet, both languages (391 tests: cpp 122, c 128, core
      detector 141) — all pass
- [x] `ruff_audit.py --ci` / `mypy_audit.py --ci` — no new findings
- [x] Direct DB verification: `godot/gdscript_vm.cpp` drops from 108 to 8 functions
      (exactly the 100 `OPCODE`-family false positives removed, all 8 remaining are
      real: `GDScriptFunction::call` etc.); `cpython/typeobject.c` confirmed zero
      `RICHCMP_WRAPPER`/`SLOT0`/`SLOT1`/`SLOT1BINFULL`/`DICT___REVERSED___METHODDEF`
      entries remaining
- [x] Both golden masters re-blessed and directly spot-checked (queried the blessed
      JSON's own `gdscript_vm.cpp` function list, confirms it matches the
      independently-verified 8-function result exactly) rather than trusting the
      terminal diff output alone, given the noted flakiness

### DIFF
--- a/docs/language_status/c.md
+++ b/docs/language_status/c.md
@@ -293,6 +293,24 @@ cpython, doom, sqlite, micropython, and more):
 | Class recall/precision | **100%** (61/61) | 100% (61/61) | 100% (61/61) | fully reconciled after this pass's fixes — see below |
 | Args exact-match | **100%** (1710/1710) | 99.9% (1709/1710) | 100% (1710/1710) | tied for best |
 
+**Update (2026-08-21):** the credit_tools adjustment this table's own GitGalaxy 100% row relied on
+(see the note directly below) is no longer needed at all — GitGalaxy now finds `slot_mp_ass_subscript`/
+`slot_nb_inplace_power`/`slot_tp_repr`/`slot_tp_hash` (and every other real function) WITHOUT ALSO
+producing the `RICHCMP_WRAPPER`/`SLOT0`/`SLOT1`/`SLOT1BINFULL`/`DICT___REVERSED___METHODDEF` false
+positives that used to sit alongside them, since `_slice_by_braces` in `detector.py` now scans
+each file for `#define NAME(...)` function-like macro definitions and excludes any match whose
+name is a known macro (the same fact universal-ctags itself already uses — confirmed via a direct
+`ctags -f -` run that ctags tags a macro name once, at its own `#define` line, and never re-tags an
+invocation of it). This was originally built to fix an analogous cpp false positive
+(`OPCODE(m_op) { ... }`, a bytecode-dispatch macro in godot/gdscript_vm.cpp) and turned out to fix
+this C shape too, for free, since both languages share the same C preprocessor and the same
+`_slice_by_braces` code path. Fresh numbers post-fix: GitGalaxy 1719/1719 = **100%** raw (no credit
+adjustment needed), ctags 1712/1733 = 98.8%, tree-sitter 1702/1790 = 95.1% — GitGalaxy now wins
+this panel outright with cleaner ground truth, not a ledger override. See
+`docs/self_scan/tri_comparison_ledger.json`'s `c/function/existence/agree[gitgalaxy,tree_sitter]_vs[ctags]`
+entry for the current verdict and the small (3-occurrence) residual finding this surfaced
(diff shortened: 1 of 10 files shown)


squid-protocol:
Follow-up commit: the \`tree-sitter-accuracy-audit\` CI check (a separate, 2-tool "tree-sitter is ground truth" system, distinct from the tri-comparison ledger discussed above) flagged \`found_functions\` regressions for cpp/c after this fix. Confirmed this is a real but *expected* consequence, not a new bug: tree-sitter's own C/C++ grammar parses \`OPCODE(x) { ... }\`/\`RICHCMP_WRAPPER(...)\` as a real function just as readily as a naive regex would (verified \`real_functions\`, i.e. tree-sitter's own parse, is byte-identical before/after this fix). GitGalaxy correctly stopped agreeing with that shared false positive, which this tool's narrow methodology reads as a recall loss against ground truth — but the ground truth itself is wrong here, the same "ground truth can be wrong" case this tool's own docstring already documents several times over (Flow-typed JS, Cython scope loss, csharp's parse-error cascade). Re-blessed both baselines and added a new dated `SCOPE & LIMITATIONS` entry explaining it, rather than treating the CI red as a real regression.

PR #2458, opened 2026-08-29
docs(tri-comparison): mandatory recall audit — every GG non-detection assessed & logged

squid-protocol:
## What

Adds **step 2.6** to the `tri-comparison-ledger-sweep` skill — a validation **gate** requiring
that every function tree-sitter OR ctags reports that GitGalaxy does not, across the whole
corpus, is individually read and sorted into exactly one of two buckets:

1. **a real GitGalaxy recall gap** → GitHub issue with an isolated repro
2. **a comparison/audit-tool artifact** → name the mechanism (macro hallucination, `#if 0` dead
   code, bodyless declaration, deliberate scope boundary, naming-convention mismatch,
   occurrence-alignment fuzz), fix the audit or document it

## Why

The ledger keys entries by discrepancy **pattern**. A shape verdict of "tree-sitter over-detects
here" can be marked `validated` **without anyone confirming every occurrence in that bucket is a
tree-sitter error rather than a GitGalaxy miss.** Confirmed real (cpp, 2026-08-29): one bucket
held ~164 occurrences — ~162 tree-sitter `OPCODE`-macro / `_FORCE_INLINE_`-parse artifacts, but
**2 genuine GitGalaxy recall gaps** (`STDAPI DllCanUnloadNow()` macro-return-type form; K&R
`PRIVATE void PlinkPrint(out,plp,tag)` untyped-param defs) that the blanket "over-detects"
verdict would have buried forever.

## Changes

- **`SKILL.md` step 2.6** — the gate, how to run it, the two buckets with per-mechanism
  resolution guidance.
- **Step 8 capstone trigger** — now requires the recall audit complete, not just a fully-
  `validated` ledger (a validated ledger with an unaudited recall gap is exactly the failure
  mode this prevents).
- **Capstone §9** must carry: the recall number, "every non-detection individually assessed",
  the bucket-1 count (each linked to its issue), and the bucket-2 per-mechanism breakdown with a
  real `file:line` each.
- **`tests/tools/recall_audit.py`** (new) — enumerates every non-detection vs tree-sitter and
  ctags at the occurrence level, with source lines, and cross-checks `measure()`'s own
  `real_functions - found_functions` so a divergence between the raw name-diff and the counted
  misses is itself surfaced.
- **`tri_comparison_README.md`** — a "is GitGalaxy missing anything?" section with a
  `RECALL_AUDIT` marker block for the standing answer.

## Follow-up (separate PR)

Executing step 2.6 for the current state: the 8 languages with func recall < 100% break down as
~3 genuine GitGalaxy gaps (shell keyword-as-argument, c K&R defs, cpp macro-return-type) + a
large tail of tree-sitter/ctags artifacts the audit is currently counting against GitGalaxy's
recall. That PR files the issues and corrects the audit.

https://claude.ai/code/session_019Zm1uVBpVEZJ9SB7bWFR8S

### DIFF
--- a/.claude/skills/tri-comparison-ledger-sweep/SKILL.md
+++ b/.claude/skills/tri-comparison-ledger-sweep/SKILL.md
@@ -370,6 +370,95 @@ lines that broke `func_start` confirmed the identical false-positive, requiring 
 same PR). A shared root cause fixed in only one of two structurally-identical rule keys is a
 half-fixed bug, not a complete one.
 
+## 2.6. The recall audit -- MANDATORY, every function GitGalaxy does not detect gets individually assessed
+
+**This is a validation GATE, not an optional deep-dive. A language's tri-comparison work is not
+"done" -- step 8's capstone must not be written -- until every function tree-sitter OR ctags
+reports that GitGalaxy does not, across the WHOLE corpus, has been individually looked at and
+sorted into exactly one of the two buckets below.** The point is a hard guarantee: *we know
+whether GitGalaxy is missing anything real, per language, and it is written down.*
+
+Why this needs to be its own step and can't be left to per-shape investigation: the ledger's
+shapes are keyed by discrepancy *pattern*. A shape like `<lang>/function/existence/
+agree[tree_sitter]_vs[ctags,gitgalaxy]` can be marked `status: validated` with a verdict of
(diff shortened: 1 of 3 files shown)


squid-protocol:
Superseded by #2463, which includes this commit plus the execution of step 2.6 (c/cpp/fortran/objc recall corrections + docs). Closing in favor of that one.

Issue #1184, opened 2026-08-09
Contiguous 'dead zones' of real functions silently lost from extraction on production files, independent of mid-file language drift


squid-protocol:
## What

On real (non-test-fixture, non-drift-flagged) production source files, Python function
extraction (`_slice_by_indentation`, Mode C) loses contiguous *ranges* of otherwise ordinary
functions while functions immediately before and after the range are found correctly. This is
distinct from #1183 (mid-file language misrouting) — none of the three files below trigger
`_partition_segments`'s language switch; every segment stays `python` throughout.

**Root cause not yet isolated** — filing this as a confirmed, reproducible pattern with concrete
evidence, not a diagnosed fix. It needs real debugging time in `_slice_by_indentation` /
`_build_indentation_safe_stream`.

## Concrete repro (ast ground truth vs. extracted, in source order)

`gitgalaxy/core/guidestar_lens.py` — first 11 functions all missing, last 4 all found:

```
line   54  MISSED  __init__
line   92  MISSED  scan_project_config
...  (9 more missed) ...
line  306  MISSED  _parse_toml_style_manifest
line  321  FOUND   _extract_execution_triggers
line  345  FOUND   _scan_gitattributes
line  403  FOUND   _scan_gitignore_evasion
line  452  FOUND   _calculate_documentation_coverage
```

`gitgalaxy/galaxyscope.py` — first 3 found, then everything from line 682 to EOF (20 functions,
including `main`) missing:

```
line   88  FOUND   execution_timeout_failsafe
line   99  FOUND   _init_worker
line  178  FOUND   _process_file_worker
line  682  MISSED  __init__
...  (19 more missed, through line 3013) ...
```

`gitgalaxy/core/prism.py` — a missing block sandwiched between two found blocks (lines 68-313
found, 425-718 missing, 731 found again) — rules out a simple "everything after one bad match
cascades to EOF" theory; whatever the mechanism is, it can apparently recover.

None of these are naming collisions or truncation (#1182) — the missed names are short,
ordinary, and unique within their file (`__init__`, `main`, `cleanup`, `unmask`, `callback`).
Precision stays high throughout (~99.7% corpus-wide once #1182's truncation is corrected for) —
this is a pure recall/loss problem, not a hallucination problem.

## Scale

Measured against GitGalaxy's own 228-file Python self-scan corpus, restricted to the 217 files
that do *not* trigger #1183's language-drift bug: 2,183 real functions (ast ground truth) vs.
[...]

PR #2506, opened 2026-08-30
core: shield string literals in positional comment stripper (#259)

### LINKED ISSUE #259: Fixed-form COBOL/Fortran comment stripping doesn't shield string literals
Every other stripping path in this file — _strip_segment_comments's generic pipeline, _strip_nested_comments — masks string literals via LITERAL_MASK_PATTERN before scanning for comment delimiters, specifically to avoid mistaking a delimiter-like character inside a string for a real comment marker. _strip_positional_comments (used for column_sensitive languages — COBOL, Fortran) does not:
pythonif "*>" in line:
    parts = line.split("*>", 1)
    code.append(parts[0]); lits.append("*>" + parts[1])
elif "!" in line:
    parts = line.split("!", 1)
    code.append(parts[0]); lits.append("!" + parts[1])
A line like PRINT *, "Warning!" gets truncated at the ! inside the string, misclassifying the rest of the line as a comment. Given COBOL/Fortran get first-class treatment elsewhere in the docs (the whole "Legacy Modernization" toolchain), this is a real gap rather than a documented tradeoff.
Fix: apply the same LITERAL_MASK_PATTERN shielding used elsewhere before the *>/! split.

squid-protocol:
## What

`Prism._strip_positional_comments` (the `positional_anchored` / `positional_abap` families — **cobol, fortran, abap**) split each line directly on `*>` / `!` / `"` with **no string-literal shielding**, unlike every other stripping path in `prism.py` (`_strip_single_line_comments`, `_strip_segment_comments`, `_strip_nested_comments` all mask via `LITERAL_MASK_PATTERN` first).

A delimiter-shaped character *inside a string literal* was misread as a real inline-comment marker:

```
_strip_positional_comments('       DISPLAY "Rate *> 5%" TO CONSOLE.\n       MOVE X TO Y.', cobol_mode=True)
  # before:  code = '       DISPLAY "Rate \n       MOVE X TO Y.'   ← "TO CONSOLE." lost, dangling unterminated "
  # after:   code = '       DISPLAY "Rate *> 5%" TO CONSOLE.\n       MOVE X TO Y.'
```

Same failure for the `!` branch (`DISPLAY "Warning!"`) and the ABAP `"` branch (`x = 'he said "hi"'`). The tail of the literal + rest of the statement moved to the comment stream, deflating `coding_loc` and silently dropping real structural signals.

The prior closure ("resolved by #273") referenced the wrong PR — #273 only touched `network_risk_sensor.py`. #1745 fixed a *different* bug in the same function (line-count geometry). No regression test guarded this.

## Fix

- Mask string/char literals **per line** (the same bounded discipline `_strip_single_line_comments` adopted for #1184) before the inline `*>` / `"` / `!` search, restoring them into both the code and comment halves after the split.
- The column-1 / column-7 anchor check stays on the **raw** line so column positions are exact.
- ABAP masks **single-quote / backtick only** — its `"` is the comment delimiter, not a string quote (new `ABAP_LITERAL_MASK_PATTERN`).
- `_mask_line_literals` gains an optional `pattern` override (backward compatible).
- **Out of scope:** a literal continued onto a later line via a fixed-form column-7 `-` indicator — masking is deliberately line-bounded (#1184). Noted in the code comment.

## Golden masters (re-blessed, both modes)

`python tests/tools/update_golden_master.py` run once per mode. **5 diffs, all on one file:** `cobol/che-che4z_lsp_project_fixtures/special_schema.cbl`, whose `DATA ... VALUE` clauses hold XML-ish string literals containing `!` (`'  <itemName>Not here!</itemName>'`). With the literal no longer truncated at the `!`:

- `Type/Safety Bypasses`: **0 → 1** (a real signal previously hidden in the mis-stripped comment portion)
- `Error & Exception Exposure`: 62.98% → 65.52% on that file, + 3 aggregate rollups

No line-shift noise anywhere else across the corpus.

## Verification

- ✅ `tests/core_engine/` (572 passed) + new `test_prism.py` cases (delimiter-inside-literal for cobol `*>`/`!`, fortran `!`, abap `"`, plus real-inline-comment control cases)
- ✅ cobol / fortran / abap extraction gauntlets + `_strict` siblings (412 passed)
- ✅ `tests/tools/crucible_check.py` — both modes PASS after re-bless
- ✅ `tri_comparison_chart.py --all --ci`, `tree_sitter_accuracy_audit.py --all --ci` (pre-existing unrelated `css` corpus-drift failure reproduces on clean `main` with these changes stashed — not introduced here)
- ✅ `ruff_audit.py --ci`, `mypy_audit.py --ci`, `dead_key_audit.py --ci`, `ruff format --check` on `prism.py`

Closes #259

### DIFF
--- a/gitgalaxy/core/prism.py
+++ b/gitgalaxy/core/prism.py
@@ -133,6 +133,14 @@ class Prism:
         # "Android/*"`, swallowed ~17 handlers). Single-line, no-escape.
         self.MULTI_STYLE_LIVE_LITERAL_MASK_PATTERN = r'("[^"\r\n]*")'
 
+        # #259: ABAP's `"` is a comment delimiter, never a string quote -- ABAP
+        # string literals are `'...'` and string templates use backtick. Masking
+        # with the shared SHIELD_PATTERN's `"..."` branch in _strip_positional_
+        # comments would swallow an ABAP comment that happens to contain an inner
+        # `"..."` pair as if it were a live literal, so ABAP's positional stripper
+        # masks single-quote / backtick literals only.
+        self.ABAP_LITERAL_MASK_PATTERN = r"((?<!\\)'(?:\\.|[^'\\])*'|(?<!\\)`(?:\\.|[^`\\])*`)"
+
         # #1271: detects a quote that opens but never closes before end-of-
         # line -- a backslash-newline-continued literal (legal in both Ruby
         # and Python) -- so _strip_single_line_comments can carry that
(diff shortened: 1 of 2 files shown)


squid-protocol:
The failing `tree-sitter-accuracy-audit` / `css` check is **pre-existing and unrelated** to this PR — `real_functions 24→25` drift from a stale css baseline (last regenerated in #2478; #2500's audit-reader changes weren't re-blessed). Verified it reproduces on `main` (`4cc5001f`) with this branch's changes stashed, and #2500 merged with the identical failure. This PR only touches `_strip_positional_comments` (cobol/fortran/abap), which cannot affect css.

Tracked separately in #2507 (also covers the `@layer`/`@media`/`@supports`-as-function FP the same audit surfaces).

All other checks green: crucible-audit both modes, full-suite-matrix ×40, tri-comparison, ruff/mypy/dead-key.

PR #2628, opened 2026-09-01
fix(swift): capture single-letter imports

### LINKED ISSUE #2543: swift: _dependency_capture requires 2+ char module names, silently dropping single-letter imports from the DAG
## Summary
Found by the #1096 control corpus: swift's `_dependency_capture` requires module names of **2+ characters** — `([a-zA-Z_][\w.]+)` — so `import a` captures nothing, the dependency edge is lost, popularity stays 0, and the Contextual Baseline Fix never fires for the target file.

## Evidence
The corpus's swift chain (main→a→b→c via `import a` etc.) is the only brace-family folder whose orphan→api conversion never fires (orphaned_logic stays 3 on all imported files); every sibling language (`use a`, `import a`, `#include "a.c"`, …) resolves and converts. Other languages' capture groups use `*`-shaped tails; swift's `+` looks unintentional.

## Impact
Small in the wild (single-letter modules are rare) but it silently truncates DAGs where they exist; the fix is one character (`[\w.]+` → `[\w.]*`).

## References
#1096 corpus ledger `swift-import-capture-min-two-chars`.

kanishk083:
### Description

  Swift's `_dependency_capture` required an initial identifier character followed by one or more
  `[\w.]` characters. This made module names at least two characters long, so a valid single-letter
  import such as `import a` was silently omitted from the dependency DAG.

  This changes the tail quantifier from `+` to `*`, preserving existing imports while accepting
  single-letter module names.

  ### Changes

  - Accept single-letter Swift module names.
  - Add `import a` as a regression case.
  - Ensure bare `import` remains invalid.
  - Add adversarial ReDoS and scaling coverage.

  ### Validation

  - Swift extraction and strict suites: **130 passed**
  - Real compiled-regex candidates: **4/4 passed**
  - Scaling sweep: linear behavior
  - AST accuracy audit: passed
  - `git diff --check`: passed
  - Full-suite run: **7,146 passed**, 14 skipped, 2 deselected, 9 xfailed, 3 xpassed, and 15 failed
  - The dependency-related failures passed after installing their expected optional dependencies: **31
  relevant tests passed**
  - The two remaining failures are Windows-host constraints involving CP1252 console output and
  symlink privileges
  - No golden-master or baseline files changed

  ### Core Engine Modification Checklist

  - [ ] **Golden Master Verification:** Both Crucible engine scans completed across 2,817 files, but
  the Windows wrapper failed afterward while printing a Unicode timing glyph. Linux CI is
  authoritative.
  - [ ] **Tri-Comparison Audit:** Requires `ctags`, which was unavailable locally; handled
  automatically by CI.
  - [ ] **Tree-Sitter Accuracy:** Local execution was blocked by Windows corpus path handling; Linux
  CI is authoritative.

  ### Proof

  - [Regex fix](https://github.com/kanishk083/gitgalaxy/blob/fix/swift-import-regex/gitgalaxy/
  standards/language_standards/languages/swift.py#L198-L201)
  - [Regression and ReDoS tests](https://github.com/kanishk083/gitgalaxy/blob/fix/swift-import-regex/
  tests/extraction/languages/test_swift.py#L235-L275)
  - [Control corpus](https://github.com/squid-protocol/language-crucible), pinned at `v1.2.0`

  ### Cross-repo

  No companion code change is required. The control-corpus reproducer already exists. Rerun the Swift
  corpus ledger after this engine change lands.

  Fixes #2543

### DIFF
--- a/gitgalaxy/standards/language_standards/languages/swift.py
+++ b/gitgalaxy/standards/language_standards/languages/swift.py
@@ -196,7 +196,7 @@ DEFINITION: dict[str, Any] = {
         # 24. import (Dependency Inclusions)
         "import": re.compile(r"^[ \t]*(?:@_exported[ \t]+)?import\s+[a-zA-Z_]\w*", re.M),
         "_dependency_capture": re.compile(
-            r"^[ \t]*(?:@_exported[ \t]+)?import\s+(?:(?:typealias|struct|class|enum|protocol|let|var|func)\s+)?([a-zA-Z_][\w.]+)",
+            r"^[ \t]*(?:@_exported[ \t]+)?import\s+(?:(?:typealias|struct|class|enum|protocol|let|var|func)\s+)?([a-zA-Z_][\w.]*)",
             re.M,
         ),
         # 25. ownership (Authorship Metadata)
--- a/tests/extraction/languages/test_swift.py
+++ b/tests/extraction/languages/test_swift.py
@@ -235,10 +235,12 @@ def test_swift_class_start_pathological(payload, expected_name):
 # ==============================================================================
 DEPENDENCY_CASES: dict[str, Any] = {
     "valid": [
+        ("import a", "a"),  # issue #2543: single-letter module
         ("import Foundation", "Foundation"),
         ("@_exported import UIKit", "UIKit"),
     ],
     "invalid": [
+        "import",
         "var importData = true",
     ],
     "pathological": [
@@ -267,3 +269,7 @@ def test_swift_dependency_capture_pathological(payload, expected_path):
     assert_pathological_dependency_match(
         SWIFT_RULES["_dependency_capture"], payload, expected_path, "swift._dependency_capture"
     )
+
+
+def test_swift_dependency_capture_redos_immunity():
+    assert_redos_immune(SWIFT_RULES["_dependency_capture"], "import " + "a." * 50000, timeout_sec=3.0)


squid-protocol (approved):

kanishk083:
The Rosetta failure is the expected result of this fix: `a.swift` API 0→3, `b.swift` 1→4, and
  `c.swift` 0→3.

  Would you prefer that I open the companion `keyword-rosetta` re-baseline PR, or will you handle it
  with the pin update?

  The Crucible/tree-sitter/tri-comparison failures occurred before testing because
  `LANGUAGE_CRUCIBLE_REF` expanded to an empty value.

squid-protocol:
Thanks for this! Your diagnosis is correct on both counts, and
the second one is our bug, not yours.

**The Crucible / tree-sitter / tri-comparison failures are ours.** You're right that
`LANGUAGE_CRUCIBLE_REF` expanded to empty. The variable is set (`v1.2.0`), but GitHub doesn't
expose repository variables to `pull_request` runs from forks, so those four workflows run
`git clone --branch ""` and die before your code is ever exercised. `rosetta-audit` happens to
guard this with `|| 'main'`; the crucible-family workflows don't. Yours is the first fork PR
this repo has had in a long while, which is why nobody had hit it. We're fixing that in a
separate PR and adding a note to CONTRIBUTING.md so the next contributor isn't left guessing.
Nothing for you to do there.

**The Rosetta failure is real and is the gate working as intended.** Your read is exactly right:
this is case 2 in `docs/self_scan/BUMPING_THE_ROSETTA_PIN.md` — an intentional, corpus-visible
improvement. The corpus plants single-letter module names deliberately, and `data/swift/expected_signals.json`
locks in the old behaviour ("`import a` never resolves and popularity stays 0"). Once the chain
resolves, orphan→api conversion fires and you get exactly the 0→3 / 1→4 / 0→3 you observed.

Nice symmetry worth noting: the corpus is what *filed* #2543 in the first place. Ledger entry
`swift-import-capture-min-two-chars` says "the `+`-quantifier looks unintentional next to other
languages' `*` shapes." Your PR is what lets that entry flip to `still_reproduces: false`.

**Yes please, go ahead and open the companion `keyword-rosetta` PR** — that'd be great. Pointers:

- Follow the corpus repo's `docs/GATING.md`: update `data/swift/expected_signals.json`, flip
  `swift-import-capture-min-two-chars` to `still_reproduces: false`, and refresh the `notes`
  field (it currently states the 2+ char rule as an invariant).
- Set the committed `ENGINE_REF` file to `pull/2628/head`. That points the corpus gates at *this*
  unmerged branch so your corpus PR goes green immediately, rather than against `main` where the
  fix doesn't exist yet.
- **One timing note:** corpus PR #20 is currently in flight and still holds `ENGINE_REF: pull/2634/head`.
  We'll land that first so you don't collide on that file — we'll ping you here when it's clear.
- Restoring `ENGINE_REF` to `main` and bumping `KEYWORD_ROSETTA_REF` on this side both need repo
  admin, so leave those to us. We'll handle the pin bump once your corpus PR merges, and
  `rosetta-audit` should go green here on the rerun.

The fix itself looks good — the `import` invalid case and the ReDoS coverage were the right
things to add. Thanks for the unusually careful writeup, and for catching our CI defect on the
way through.

squid-protocol:
Correction to my note above: **the coast is already clear — please go ahead now, no need to wait.**

I said corpus PR #20 was in flight holding `ENGINE_REF` and that we'd land it first. That was
wrong. keyword-rosetta#19 had already merged the same re-baseline, corpus `main` is back to
`ENGINE_REF: main`, and the pin here already points at it. #20 is a leftover duplicate on our
side and we'll close it — it never had a claim on `ENGINE_REF`.

So nothing is blocking you: open the corpus PR whenever you like, with `ENGINE_REF` set to
`pull/2628/head`. Sorry for the misdirection.

Separately, the four Crucible/tree-sitter/tri-comparison failures now have a fix in #2637 — it
replaces the bare `fatal: Remote branch  not found` with an explicit message saying the pin is
unavailable on fork PRs and that it isn't your change. Thanks again for surfacing it.

kanishk083:
Thanks ,I checked the current state before opening the companion PR. Since your message, gitgalaxy #2641 and keyword-rosetta #24 have landed. The Swift regex fix, corpus re-baseline, and ledger resolution are now on `main`, and #2543 is closed.

PR #2628 still carries two test-hardening additions that are absent from current `main`: the bare `import` negative case and the dedicated ReDoS scaling guard.

Would you prefer that I rebase and reframe #2628 as a test-only hardening follow-up, or close it as superseded by #2641?

squid-protocol:
Thanks for this, @kanishk083 — the diagnosis was correct and the fix was the right one-character change.

Unfortunately it got overtaken. #2641 (`5813a9cb`, merged Sep 1) landed the byte-identical `[\w.]+` → `[\w.]*` change as part of a wider Rosetta corpus sweep, along with an `("import a", "a")` regression case tagged to the same issue (#2543). So the swift.py hunk here is now a no-op against `main`, and merging as-is would leave a duplicated test entry in `DEPENDENCY_CASES["valid"]`.

Two things from this PR were **not** covered by #2641, and I've opened #2666 to add them, with you as `Co-Authored-By`:

- `"import"` as an explicit invalid case — relaxing the tail to `*` is exactly the change that could go one step too far and match a bare `import`, and nothing pinned that it doesn't.
- `test_swift_dependency_capture_redos_immunity` — `func_start` and `args` already had ReDoS coverage, but `_dependency_capture` did not, even though its dotted-module tail is the one unbounded quantifier in the rule.

For the record, none of the red checks here were your fault:

- The four corpus-backed audits (`crucible-audit` ×2, `tree-sitter-accuracy-audit`, `tri-comparison-audit`) failed because GitHub withholds the `LANGUAGE_CRUCIBLE_REF` repository variable from fork PRs, so the corpus clone got an empty ref. That's fixed as of #2665 — these audits now fall back to the in-repo pin and run properly on fork PRs.
- `rosetta-audit` failed against a keyword-rosetta pin that expected `api 0/1/0` for the swift fixtures. The corpus was re-baselined for #2641 and the pin bumped, and it now expects exactly the `3/4/3` your branch produced.

Closing as superseded, but the analysis in your description was sound and the ReDoS/scaling work is what made the salvaged coverage worth keeping. Thanks again.

kanishk083:
Thanks for salvaging the additional coverage and for the detailed explanation. Glad the contribution also helped expose the fork-CI gap.

Issue #2675, opened 2026-09-02
safety_bypasses/io over-claim: four rules count tokens another rule owns (livecode global, lua _G+collectgarbage, fortran COMMON, dockerfile apt-get clean) — the #2659 shape


squid-protocol:
Batch B.4 of #2669. Four post-Batch-A residual cells on `io` / `safety_bypasses` turn out to be one shape, and all four are regex-level — **B.3 material**. Corpus-side residuals from the same triage go to keyword-rosetta separately.

## The shape

A rule claims a token that **another rule already owns**, or matches a token in a form too broad to mean what the rule means. Same family as #2659 (`doc` claiming what `ownership` owns) and #2544 (a guard that does not guard).

Every one of the four lands **exactly on the planted value**, and in every case the token removed is still counted by its owning rule, which stays at its own planted value.

| language | signal | now | fixed | planted | token claimed | owned by |
|---|---|---|---|---|---|---|
| livecode | `safety_bypasses` | 4 | **2** | 2 | `global\s+` | `globals` |
| lua | `safety_bypasses` | 4 | **2** | 2 | `_G`, `_ENV` / `collectgarbage` | `globals` / `cleanup` |
| fortran | `safety_bypasses` | 3 | **2** | 2 | `COMMON` | `globals` |
| dockerfile | `io` | 5 | **3** | 3 | bare `apt-get`/`apk`/`yum`/`dnf` | — (over-broad form) |

## 1. livecode `safety_bypasses` claims `global`

```py
r"\b(disable\s+messages|unlock\s+(?:screen|messages)|global\s+)\b|\bdo\s+(?![a-zA-Z_]\w*\b)"
```

`disable messages`, `unlock screen` and dynamic `do` are genuine bypasses. `global ` is a **scope declaration**, and `globals` owns it exclusively (`\b(global\s+|the\s+global|…)\b`). The corpus's `probe_globals` plants two `global` declarations in `a.lc`, and each scores `safety_bypasses` as well — the entire +2.

Fix: drop the `global\s+` alternative.
```py
r"\b(disable\s+messages|unlock\s+(?:screen|messages))\b|\bdo\s+(?![a-zA-Z_]\w*\b)"
```

## 2. lua `safety_bypasses` claims `_G`/`_ENV` and `collectgarbage`

```py
r"\b(rawget|rawset|rawlen|debug\.[a-zA-Z0-9_]+|collectgarbage|_G|_ENV|getfenv|setfenv)\b"
[...]

Issue #2671, opened 2026-09-02
apex import rule: re.IGNORECASE neutralises the [A-Z] guard, so every qualified reference (foo.bar) counts as an import


squid-protocol:
Found by the Batch A regen in keyword-rosetta (gitgalaxy#2669 A.5), while attributing a new red cell: apex `import` read 6 against a cross-language median of 3.

## The defect

`LANGUAGE_DEFINITIONS["apex"]["rules"]["import"]` is:

```
\bType\.forName\b|(?!(?:System|Database|Schema|Auth|Cache|Chatter|EventBus|Limits|Messaging|RestContext|Test)\b)\b[a-zA-Z_]\w*\.[A-Z]\w*\b
```

compiled with `re.IGNORECASE`. The second alternative is clearly meant to catch a *type* reference — lowercase receiver, **capitalised** member — but `re.IGNORECASE` applies to the whole pattern, so `[A-Z]` matches any letter and the alternative degrades to "any `word.word`". Every method call in an Apex file is therefore counted as an import.

Micro-repro (live registry, engine 3ac91822):

```
foo.bar          -> ["foo.bar"]      # should not match
conn.clear       -> ["conn.clear"]   # should not match
Logger.info      -> ["Logger.info"]  # should not match
System.debug     -> []               # correctly excluded by the lookahead
Database.query   -> []               # correctly excluded by the lookahead
```

Recompiling the identical pattern **without** `re.IGNORECASE` returns `[]` for all three of the first group, which isolates the flag as the whole cause.

Scanned every language: apex is the only rule that combines `re.IGNORECASE` with an `[A-Z]` guard, so the blast radius is apex alone.

## Why it surfaced now

The keyword-rosetta corpus expresses the SPEC dependency chain main->a->b->c as `Type.forName(...)`, planted in #2669 A.3 (apex previously had no chain at all). That is +1 legitimate import per planting file. The corpus also contains ordinary Apex code — `Logger.info`, `Log.error` in `b.cls`, `conn.clear` in `c.cls` — and those three incidental method calls are counted as imports too:

| file | matches |
|---|---|
| main.cls | `Type.forName` |
| a.cls | `Type.forName` |
| b.cls | `Type.forName`, `Logger.info`, `Log.error` |
| c.cls | `conn.clear` |

= 6 against a planted intent of 3. Before A.3 the count coincidentally read 3 (the three incidental hits with no chain), which is why nothing caught it earlier — the number was right for the wrong reason.

[...]

Issue #2670, opened 2026-09-02
perl doc: a correctly-closed POD block counts twice (=pod + =cut) — the #2658 delimiter shape


squid-protocol:
Found while fixing the keyword-rosetta corpus's perl shell (gitgalaxy#2669 A.3).

## The rule

```py
"doc": re.compile(r"^=(?:pod|head[1-6]|item|over|back|cut|begin|end|encoding|for)\b", re.M)
```

`=cut` is in the same alternation as the openers, and nothing pairs them. A **correctly written**
POD block therefore always counts **at least 2** doc hits:

```perl
=pod
some documentation
=cut          # <- counted as a second, independent doc hit
```

This is exactly the shape #2658 fixed for python (`"""` opener and closer counted separately with
no pairing) — resolved there by matching the full delimited span as one bounded hit.

## Differential (real scan, full_precision, engine main @ 3ac91822)

keyword-rosetta `data/perl/main.pl`, one line added and nothing else changed:

| main.pl | doc |
|---|---|
| `=pod` … (unterminated) | 1 |
| `=pod` … `=cut` | **2** |

Every other recorded signal is identical across the two scans.

## Why it is not just a counting nit

The corpus's perl shell had been left with an **unterminated** `=pod` — the only way to get
`doc = 1` — and that turned out to be masking a much larger effect, because
`detector.py` (~L3042, the #2239 POD strip) blanks from a POD opener to the next `=cut`, or to
**end of file** when there is none:

```py
m_end = pod_end.search(safe_code, m_start.start())
end_idx = m_end.end() if m_end else len(safe_code)
```

So the whole of `main.pl` was blanked for structural scanning. Closing the block restores it:

| main.pl | before (`=pod` unterminated) | after (`=pod … =cut`) |
|---|---|---|
| `avg_func_loc` | 0.0 | 6.75 |
| `avg_func_complexity` | 0.0 | 0.75 |
| `func_internal_density` | 0.0 | 0.111 |
| `structural_mass` | **104.52** | 11.9 (siblings: 8.3 / 11.44 / 9.36) |

Corpus-wide that was `functions_found` 9 vs a 13 median for perl, and it is the direct cause of
perl's depressed `avg_func_loc` (-27%), `avg_func_complexity` (-38%), `max_func_complexity`
(-50%), `func_complexity_gini` (-53%), `func_internal_density` (-49%) and `control_flow_ratio`
(-40%) cells, plus `structural_mass` +212%.

[...]

Issue #2661, opened 2026-09-02
cobol: doc rule's AUTHOR-vs-ownership collision needs a corpus-side doc line planted before the engine fix (#2659 follow-up)


squid-protocol:
Follow-up from #2659 (livecode + agc_assembly + fortran + html fixed there; cobol deliberately excluded).

## What's different about cobol

cobol's doc rule shares the identical bare-colon-tag collision with AUTHOR/ownership that #2659 fixed in four other languages: `^(?:...)?[ \t]*(?:AUTHOR|DATE-WRITTEN|DATE-COMPILED|REMARKS|INSTALLATION)\.|\*>\s*@(?:param|return|author)`. AUTHOR overlaps cobol's ownership rule (`^(?:...)?[ \t]*AUTHOR\.\s+([^\n]+)`) exactly the same way.

But cobol's rosetta corpus shell (`data/cobol/main.cbl`) has **no other doc-idiom content** — its only doc-matching line is `AUTHOR. KEYWORD-ROSETTA GENERATOR.`. Removing AUTHOR from doc (the same fix applied elsewhere) would drop cobol's doc signal from 2 to **0**, not 1 — breaking the corpus's planted intent (SPEC.md wants exactly 1 doc-comment per language) rather than deduping a double-count. That's a materially different consequence than the four already-fixed languages, which all had a separate genuine doc-idiom line already present.

## Fix shape

Two-sided, needs sequencing:
1. **Corpus-side first**: plant a genuine cobol doc-idiom line in `main.cbl` independent of AUTHOR — cobol's own rule already supports the structured form (`*> @param`/`*> @return`/`*> @author` via COBOL 2002+ free-format inline comments), so add one, e.g. `*> @return the probe result`.
2. **Then engine-side**: remove AUTHOR from cobol's doc rule's bare alternative (same one-line change as #2659), leaving DATE-WRITTEN/DATE-COMPILED/REMARKS/INSTALLATION plus the `*>` structured form.
3. Corpus manifest doc stays at 1 throughout (2→1 pre-corpus-fix isn't needed since step 1 keeps it at the planted intent — verify the exact sequencing empirically before merging either side).

Found via the #2659 dispatch's cross-language Author/doc audit; excluded from that fix to avoid landing an engine change with an unverified corpus-breaking side effect.

squid-protocol:
[...]

PR #2681, opened 2026-09-02
fix(core-engine): Rosetta batch 3 wave 2 — doc-comment pairing, rule over-claims, three None rules (#2669)

### LINKED ISSUE #2672: doc rules count one doc comment once per marker AND once per tag — 18 languages, 8 red rosetta cells (the #2658 shape, unswept)
Batch B.1 of #2669 (child of epic #2560). Filed by a T1 session; implementation is B.3 (T2), bless is B.5 (T1).

## The defect

A `doc` rule that lists **both** a doc-comment *marker* (`/**`, `///`, `=pod`) **and** the doc *tags*
that live inside that marker's block (`@param`, `@return`, `- parameter`, …) as independent
alternatives counts one doc comment once per marker **plus** once per tag.

This is exactly the shape fixed for python/embedded_python in **#2658** (paired `"""` delimiters
into one bounded span). It was never swept across the other languages that carry it.

Post-Batch A (keyword-rosetta `357345f`, the pin `KEYWORD_ROSETTA_REF` currently points at), the
rosetta corpus plants **exactly one** doc construct per language, so `doc` should be 1 everywhere.
Eight languages read 2:

| language | planted construct | matches today |
|---|---|---|
| java | `/** @param argv probe input */` | `/**` + `@param` |
| groovy | `/** @param argv probe input */` | `/**` + `@param` |
| kotlin | `/** @param argv probe input */` | `/**` + `@param` |
| php | `/** @param $argv probe input */` | `/**` + `@param` |
| scala | `/** @param argv probe input */` | `/**` + `@param` |
| solidity | `/// @param flag the probe input` | `///` + `@param` |
| perl | `=pod … =cut` | `=pod` + `=cut` (**#2670**) |
| cobol | `AUTHOR.` + `*> @return` | `AUTHOR.` is claimed by `doc` **and** `ownership` (**#2661** step 2, the **#2659** shape) |

Verified by running each language's live `doc` rule over the real `prism.split_streams()` output
for its corpus files — the raw match lists above reproduce `docs/bias_data.json` `doc` exactly
(8 languages at 2, the other 38 at 1). `doc` is counted in both `coding_analysis` (per segment, over
`code_stream`) and `comment_analysis` (over `comment_stream`), both `len(finditer(...))`, so
non-overlapping alternation is what decides the count.

## It is much worse off-corpus

The corpus only plants **one** of {marker, tag} in the other family members, which is the sole
reason they read 1 — not because their rule is right. One ordinary Doxygen/JSDoc/ApexDoc block:

```c
/** @brief do it
 * @param x in
 * @return out
 */
```

| language | `doc` for that one comment |
|---|---|
| c, cpp, objective-c, apex | **4** |
| java, javascript, typescript | **3** |
| dart, yacc | **2** |
| swift (`/// - parameter` + `/// - returns:`) | **4** for a 2-line comment |
| solidity (`/// @notice` + `/// @param`) | **4** for a 2-line comment |

So this is a real-repo scoring defect, not a corpus artifact: every well-documented C/C++/Java/TS
file has been scoring `doc` proportional to *tag density* rather than *doc-comment count*. Since
`doc` is a **defense** term in `_calc_documentation`
(`gitgalaxy/metrics/signal_processor.py` ~L1499, `doc_weight` 1.0, alongside `doc_loc` at 0.33),
the effect is a silently deflated `risk_documentation` for tag-heavy code. Documentation *mass* is
already carried by `doc_loc`; `doc` counting comments once is the coherent semantic, and it is the
one #2658 already established for python (a docstring with three `:param:` tags counts 1 today).

**Correct precedent already in the registry** — ada, assembly, haskell and livecode fuse marker and
tag into a single alternative (`--[ \t]*(?:@param|@return|…)`), so they cannot double-count. css and
sqlite are safe for the same reason. rust and zig list markers only, no tags.

## Fix shape

Per language, in this alternation order:

1. **Block form first** — `/\*\*[\s\S]{0,15000}?\*/` (and objective-c's `/\*![\s\S]{0,15000}?\*/`),
   so the whole block including its tags is one hit. Identical bound and non-greedy body to #2658.
2. **Line-marker form next** — `///` → `///[^\n]*`, so `/// @param x` is one hit, not two.
3. **Bare tags last**, unchanged, so a tag *outside* any doc comment still counts (java's `@Operation`
   / `@Schema` are real annotations in `code_stream`, not javadoc).
4. **Remove the tag that `ownership` already owns** (the #2659 rule): drop `AUTHOR` from cobol's
   `doc` header alternation, and `@author` from apex's.

perl (#2670) takes the same treatment in POD vocabulary — pair the opener to `^=cut`:

```
^=(?:pod|head[1-6]|item|over|back|begin|end|encoding|for)\b[\s\S]{0,15000}?^=cut\b
|^=(?:pod|head[1-6]|item|over|back|begin|end|encoding|for)\b
```

apex needs one extra half-step. Its `doc` and `ownership` both list `@author`, but `ownership`
requires a colon (`@author:\s+`), so idiomatic ApexDoc `@author Joe` is doc-only (doc 3, ownership 0)
while `@author: Joe` double-counts. Dropping `@author` from `doc` alone would lose the bare form
entirely, so relax that one alternative to colon-optional, leaving the prose-risky
`Author|Copyright|…` alternatives colon-required:

```
(?:@author:?\s+|(?:Author|Created by|Maintainer|Copyright|Tim Berners-Lee):\s+)([^\n]+)
```

Checked: `@author Joe` → doc 1 / own 1; `@author: Joe` → doc 0 / own 1; the corpus's
`// Author: …` → doc 1 / own 1 (unchanged); prose `the Author of this module` → 0 / 0.

**Explicit non-goal.** A run of consecutive `///` lines still counts once *per line* (csharp and rust
score 2 for a 2-line `///` comment before and after). Folding runs into one comment would move every
`///` language with no corpus signal to validate against, and it is a separate semantic decision —
leave it, and ledger it if it ever surfaces as a cell.

## Prototype result

Candidate patterns for all 18 family members (the 8 above plus c, cpp, objective-c, dart, javascript,
typescript, swift, yacc, csharp) run over the full 46-language corpus through `prism.split_streams`,
both streams:

- **Exactly 8 languages move**, all 2 → 1: cobol, groovy, java, kotlin, perl, php, scala, solidity.
- **36 languages unchanged.** No `doc` cell moves anywhere else in the corpus.
- Every language then reads the planted `doc` = 1 — the whole `doc` column goes green.
- Idiomatic-snippet recheck: c/cpp/objective-c/apex 4 → 1, java/javascript/typescript 3 → 1,
  dart/yacc 2 → 1, swift 4 → 2 and solidity 4 → 2 (two `///` lines, per the non-goal above).

ReDoS detonation on the new quantified alternatives, 200k+ chars of never-terminated input:
`java` candidate 0.0062s, `solidity` candidate 0.0062s, `perl` candidate 0.0020s. Safe — bounded and
non-greedy, same profile as #2658's probe. Unterminated blocks now fail closed (no match) instead of
counting the opener, which is #2658's accepted trade-off.

Caveat worth carrying into B.3: `comment_stream` is a concatenation of a file's comments, so a
malformed unterminated `/**` can pair with a *later* block's `*/` within the 15000-char bound. Same
property python has had since #2658; bounded, and only reachable on malformed input. For perl the
POD hits land in `code_stream`, which is segmented — pairing assumes a POD block does not straddle a
segment boundary.

## Scope for B.3 / B.5

- Registry-regex layer only. No slicer, DAG or scoring change — keeps #2669's one-layer-per-PR rule.
- Strict tests in `tests/extraction/languages/test_<lang>_strict.py` (all 18 targets already have a
  `_strict` file except c, cpp, csharp, dart, objective-c, swift, typescript, yacc — those get one):
  positive (one doc comment = 1), negative (a tag outside any doc comment still counts), and a ReDoS
  timing probe per new quantified pattern.
- B.5 bless: expect `risk_documentation` to **rise** on real repos with tag-heavy doc comments — the
  defense term was inflated. That is the intended direction, and the golden-master diff should show
  it only where `doc` moved.
- Closes-with: **#2670** (perl) and **#2661** step 2 (cobol) are members of this family and should be
  closed by the same PR. **#2671** (apex `import` / `re.IGNORECASE`) is a *different* rule but the same
  one-regex/one-test shape and the same apex file, so B.3 should carry it in the same PR.
- Corpus side: cobol re-blesses `doc` 2 → 1, retiring ledger `cobol-2661-author-doc-collision`; perl
  retires `perl-pod-delimiter-double-count`. Both are `still_reproduces: true` today.

squid-protocol:
### Description

Batch B.3 **wave 2** of #2669 — the registry-regex fixes that DO move the control corpus. Wave 1 (#2677, merged) was the zero-movement half.

Closes #2672, #2670, #2661, #2671, #2675, #2645, #2646, #2647.

| commit | issues | effect |
|---|---|---|
| `c4a8e601` | #2672 (+#2670, #2661 step 2) | doc comment counts once, not once per marker + once per tag — 17 languages |
| `660101d7` | #2672 + #2671 | apex doc block, `@author` → ownership, import type-guard restored |
| `28de6f50` | #2675 | three rules stop counting tokens another rule owns |
| `d79b6350` | #2645, #2646, #2647 | three `None` rules wired |

One layer only (registry regex), per #2669's stacking rules.

### Corpus movement — 12 languages, each with only its intended delta

```
doc 2 -> 1   java  groovy  kotlin  php  scala  solidity  perl  cobol
import 6->3  apex                    safety_bypasses 4->2  livecode, lua
io 5 -> 3    dockerfile              (fortran withdrawn, see below)
```

Every mover lands **exactly on its planted value**. The other 12 languages this PR touches stay green: c, cpp, objective-c, dart, javascript, typescript, swift, csharp, yacc, html, yaml, fortran. `na_check --ci`: no new unreviewed absences.

The nine corpus-inert doc languages read 1 today only because the corpus plants one of {marker, tag} and never both — the defect was latent, not absent. One ordinary Doxygen block scored `doc` **4** in c/cpp/objective-c and **3** in java/javascript/typescript.

### Two things withdrawn after implementation

**fortran is withdrawn from #2675.** I filed `COMMON` in `safety_bypasses` as an over-claim. `test_fortran_strict.py:418` has documented the opposite since #1059 — *"an intentional double-classification, not a false collision"* — and it is right: a COMMON block lets distinct program units alias the same storage with no type checking, which is genuinely unchecked memory sharing, unlike livecode's scope declaration or lua's table reference. The #2650/#2659 sweep touched that same file for the author-header class and deliberately left COMMON alone. Implementation had renamed the test and inverted its assertion; that was reverted. fortran `safety_bypasses` 3 vs planted 2 is intended morphology for the ledger.

**#2644 (yacc `class_start`) is pulled** — details on its own thread. yacc is absent from `detector.py`'s `_CLASS_START_NAMED_EXTRACTION_LANGS`, so wiring its `class_start` would not use the new `%union` rule at all: it falls through to the legacy generic `class|struct|interface|trait|enum` fallback, which then claims the **57** bare `struct Foo` declarations in the embedded C action code of real grammars — measured **17 and 9 phantom classes** on the two real `.y` files, against an honest 0 today. Fixing it means editing the detector allowlist, i.e. the slicer layer, which this PR's one-layer rule excludes. The `%union` rule itself is validated (1/1 on both real files) and recorded on #2644 for its own PR.

### On the three newly-wired rules

`docs/GATING.md` makes a cell n/a **because its rule is `None`**, so wiring one ends the exemption. The corpus contains no `srcdoc`, no `author:` and no teardown verb, so all three still measure 0 — and because the manifests already record 0, `verify_language.py` and `rosetta-audit` stay green. What changes is the **bias report**: those cells leave n/a (excluded from medians) and become comparable 0s against medians of 3, 1 and 2. The companion corpus PR plants the idioms so they never surface red.

### Cross-repo

- Companion corpus PR: **keyword-rosetta#TBD** — reblesses the 12 movers, plants the three idioms, and carries keyword-rosetta#33's own work. It runs with `ENGINE_REF=pull/<this PR>/head` until this merges.
- `KEYWORD_ROSETTA_REF` bump owed in this PR once the corpus PR merges — to the **full 40-character SHA** (#2680).

### Core Engine Modification Checklist

- [ ] **Golden Master Verification** — pending; scoped diff and bless to follow in this PR.
- [x] **Tri-Comparison Audit** — to re-run post-bless.
- [x] **Tree-Sitter Accuracy** — to re-run post-bless.

Full extraction suite **6415 passed**, 3 skipped, 8 xfailed. `ruff format` clean on all 46 touched files; `ruff check` clean on every language file (test-file S101 and one RUF005 are pre-existing on main).

### DIFF
--- a/gitgalaxy/standards/language_standards/languages/apex.py
+++ b/gitgalaxy/standards/language_standards/languages/apex.py
@@ -160,7 +160,12 @@ DEFINITION: dict[str, Any] = {
             re.I | re.M,
         ),
         # 13. doc: Structured Documentation. ApexDoc annotations and metadata blocks.
-        "doc": re.compile(r"/\*\*|@description|@param|@return|@author|@date|@example", re.I),
+        # #2672: block form first (`/** ... */`, bounded/non-greedy, identical shape to
+        # #2658's python fix) so a whole ApexDoc comment -- markers plus every tag inside
+        # it -- counts once, not once per marker plus once per tag. `@author` removed from
+        # the bare-tag fallback entirely: it's owned by `ownership` (the #2659 shape); see
+        # that rule's own comment for the companion colon-optional relaxation this required.
+        "doc": re.compile(r"/\*\*[\s\S]{0,15000}?\*/|@description|@param|@return|@date|@example", re.I),
         # 14. test: Testing & Assertions. Salesforce test execution and assertion markers.
         "test": re.compile(
             r"@isTest|@TestSetup|@TestVisible|\b(?:Test\.startTest|Test\.stopTest|System\.assert|Assert\.(?:isTrue|isNotNull|areEqual)|Test\.setMock)\b",
@@ -209,8 +214,16 @@ DEFINITION: dict[str, Any] = {
         # 24. import: Dependency Inclusions.
         # Apex lacks a native import keyword. Cross-package dependencies are established via
         # reflection (Type.forName) or explicitly namespaced static invocations.
+        # #2671: the whole pattern was compiled with re.IGNORECASE (needed so
+        # `Type.forName`/`TYPE.FORNAME` both match), which also neutralised the `[A-Z]`
+        # guard meant to isolate a genuine type reference (lowercase receiver, capitalised
+        # member) -- under re.I, `[A-Z]` matches any letter, so the alternative degraded to
+        # "any word.word" and every ordinary method call (foo.bar, conn.clear, Logger.info)
+        # counted as an import. `(?-i:...)` locally turns case-insensitivity back OFF for
+        # just the member-name class, so the guard actually guards while `Type.forName` and
+        # the namespace-exclusion lookahead stay case-insensitive as before.
         "import": re.compile(
-            r"\bType\.forName\b|(?!(?:System|Database|Schema|Auth|Cache|Chatter|EventBus|Limits|Messaging|RestContext|Test)\b)\b[a-zA-Z_]\w*\.[A-Z]\w*\b",
+            r"\bType\.forName\b|(?!(?:System|Database|Schema|Auth|Cache|Chatter|EventBus|Limits|Messaging|RestContext|Test)\b)\b[a-zA-Z_]\w*\.(?-i:[A-Z]\w*)\b",
             re.I,
         ),
         "_dependency_capture": re.compile(
@@ -218,8 +231,16 @@ DEFINITION: dict[str, Any] = {
             re.I,
         ),
         # 25. ownership: Authorship indicators.
+        # #2672: `@author` requiring a colon (`@author:\s+`) meant idiomatic, colon-less
+        # ApexDoc (`@author Joe`) was never claimed by ownership at all -- it was
+        # doc-only (before this fix, double-counted there too). Now that `doc` drops
+        # `@author` entirely (see that rule's comment), the bare form would be lost with
+        # nowhere left to count it. Relaxed to colon-optional for just the `@author`
+        # alternative; the remaining prose-risky alternatives (Author|Created by|
+        # Maintainer|Copyright|Tim Berners-Lee) stay colon-required so ordinary prose
+        # ("the Author of this module") still can't false-positive.
         "ownership": re.compile(
-            r"(?:@author|Author|Created by|Maintainer|Copyright|Tim Berners-Lee):\s+([^\n]+)",
+            r"(?:@author:?\s+|(?:Author|Created by|Maintainer|Copyright|Tim Berners-Lee):\s+)([^\n]+)",
             re.I | re.M,
         ),
         # --- PHASE 4: SPECIALIZED SUB-SYSTEMS ---
--- a/gitgalaxy/standards/language_standards/languages/c.py
+++ b/gitgalaxy/standards/language_standards/languages/c.py
@@ -199,7 +199,17 @@ DEFINITION: dict[str, Any] = {
         # 12. dead_code (Commented Logic / Deprecated Trails)
         "dead_code": re.compile(r"(?://|/\*)[ \t]*(?:if|for|while|struct|union|enum|void|int|return)\b"),
         # 13. doc (Structured Documentation)
-        "doc": re.compile(r"///|/\*\*|@param|@return|@brief|@details|\\param|\\return|\\brief|\\details"),
+        # BUG FIX #2672: `/**`, `///` and the Doxygen tags (`@param`,
+        # `\param`, ...) were independent alternatives, so one Doxygen
+        # comment counted doc proportional to its tag density. Block form
+        # first (bounded 0,15000 chars non-greedy span, the #2658 shape),
+        # then the line-marker form so `/// @param x` is one hit per line,
+        # not two; bare tags stay last so a tag outside any doc comment
+        # still counts. No corpus movement -- the rosetta corpus only
+        # plants one of {marker, tag} for this language.
+        "doc": re.compile(
+            r"/\*\*[\s\S]{0,15000}?\*/|///[^\n]*|@param|@return|@brief|@details|\\param|\\return|\\brief|\\details"
+        ),
         # 14. test (Testing & Assertions)
         "test": re.compile(
             r"\b(?:TEST|TEST_F|TEST_CASE|CU_ASSERT|RUN_TEST|EXPECT_[A-Z_]+|ASSERT_[A-Z_]+)\b|\bassert\s*\("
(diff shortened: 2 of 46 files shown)


squid-protocol:
Cross-repo step 3 done: **keyword-rosetta#35 merged** (`216d57abf731477014edde452e03089d3e148f64`) and `KEYWORD_ROSETTA_REF` bumped to that full 40-character SHA. Re-running `rosetta-audit`, which should now go green — the corpus at that commit carries the reblessed manifests these fixes produce.

One item still owed on the corpus side: **#35 merged with `ENGINE_REF` still set to `pull/2681/head`**, so keyword-rosetta `main` is currently pinned to this branch rather than to engine `main`. Resetting it right now would turn that repo red, because engine `main` does not yet contain these fixes and the reblessed manifests would fail against it. So the reset follows immediately after this PR merges, as its own one-line PR. Tracked here so it is not forgotten — a stale `ENGINE_REF` on main silently builds every future corpus PR against a frozen engine ref.

PR #2699, opened 2026-09-03
ci: rosetta-audit without pins -- baseline-classified, refuses to no-op (#2682)

### LINKED ISSUE #2682: Cross-repo choreography is circular, and every way it broke today failed green — drop the ENGINE_REF dance, machine-check the invariants
Raised after running the full cross-repo choreography end-to-end for Batch B.3 (gitgalaxy#2681 + keyword-rosetta#35). Three separate things went wrong in one day, and they share a root cause.

## The choreography is circular by construction

- Corpus expected values depend on engine code → the corpus PR needs `ENGINE_REF=pull/N/head`
- The engine PR's `rosetta-audit` depends on the corpus SHA → needs the corpus PR merged
- But `ENGINE_REF` must be back to `main` before the corpus PR merges — and with `main`, the corpus gates fail, because engine `main` does not yet contain the fixes

No ordering satisfies all three. `docs/GATING.md`'s step 2 in keyword-rosetta showed the strain directly — it trailed off mid-sentence: *"engine PR must merge before the reset lands... in practice:"*. I rewrote that paragraph to read cleanly earlier today, which made a procedure that does not close **sound** like it does. That edit should be revisited.

In practice the deadlock is broken by merging the corpus PR with a knowingly-wrong `ENGINE_REF`, which is what happened: keyword-rosetta#35 merged with `ENGINE_REF=pull/2681/head`, so corpus `main` is now pinned to an engine branch and owes a reset that nothing enforces.

## The dance is not necessary

**Neither repo requires any status check to merge:**

```
gitgalaxy main:        required_status_checks.contexts: []   checks: []
keyword-rosetta main:  "Branch not protected"
```

So `rosetta-audit` is advisory. The entire `ENGINE_REF` mechanism exists to make an advisory check green *before* merge. A simpler order has always been available:

1. Merge the engine PR — `rosetta-audit` red, advisory, harmless
2. Merge the corpus PR — `ENGINE_REF` never leaves `main`; its gates pass because engine `main` now has the fixes
3. Bump `KEYWORD_ROSETTA_REF`

No `ENGINE_REF` edit, no reset debt, no deadlock, one fewer thing to forget.

`ENGINE_REF`'s real value — verifying corpus numbers against unmerged engine code *before* committing to them — does not require a committed file. It is already achievable locally, and was: pointing `GALAXYSCOPE_BIN` at the engine working tree ran all 46 gates green before anything was pushed.

## Everything that went wrong today failed GREEN

| failure | how it presented |
|---|---|
| `KEYWORD_ROSETTA_REF` set to a 7-char SHA (#2680) | `actions/checkout` resolved it as a ref name, job died at checkout — indistinguishable from a real gate failure, red for hours on main |
| `ENGINE_REF` left at `pull/N/head` on corpus main | nothing fails; future corpus PRs would silently build against a frozen ref |
| `tree_sitter_accuracy_audit.py` / `tri_comparison_chart.py` without `galaxyscope` on PATH (#2680) | skipped every language, printed `all OK`, exit 0 |

Three cross-repo references that were wrong, none of which announced itself. Documentation caught none of them — including the documentation written the same day.

## Proposals, cheapest first

1. **Drop `ENGINE_REF` from the normal flow.** Document merge-engine-then-corpus-then-pin as the default, and keep `ENGINE_REF` only for the rare case where the corpus genuinely must be verified in CI against unmerged engine code. Removes the deadlock and the reset step entirely.
2. **Assert the invariant on push to corpus main.** A one-step workflow that fails (or opens an issue) when `ENGINE_REF != main` on `main`. Catches the forgotten reset automatically instead of relying on a person.
3. **Make the pin format machine-checked.** A step in `rosetta-audit.yml` that fails loudly when `KEYWORD_ROSETTA_REF` is not 40 hex characters, before the checkout step swallows it as a ref name.
4. **Make "did it actually run?" explicit.** Every audit that can no-op should assert a plausible unit count and fail otherwise — `0 languages checked, all OK` must be a failure, not a pass.

(2) and (3) are each a handful of YAML lines and would have caught two of today's three incidents outright.

squid-protocol:
### Description

Closes #2682. Makes the keyword-rosetta gate behave the way the tri-comparison system already does: a PR-time audit that only measures and is baseline-gated, no pin on either side, and a push-to-main job (in the corpus repo) that regenerates the chart by itself.

**Root cause, in one paragraph.** `ENGINE_REF` (corpus) and `KEYWORD_ROSETTA_REF` (here) pointed at each other. The corpus PR needed `ENGINE_REF=pull/N/head` to pass, had to be back on `main` before merging, then failed against `main` because the engine PR was not in it yet, while the engine PR waited for a corpus SHA that only existed after that merge. In practice the corpus merged with a knowingly wrong ref and owed a reset nothing enforced (keyword-rosetta#39 was a PR whose only content was that reset). Neither repo requires a status check to merge, so both pins were only ever making an advisory check green before a merge that happened anyway.

### What changes

| | |
|---|---|
| `.github/workflows/rosetta-audit.yml` | Checks out keyword-rosetta at `main`; no variable. Builds the PR's base branch in a second venv and calls the script below. `rosetta:rebless-owed` label → regressions become warnings. |
| `tests/tools/rosetta_audit.py` (new) | Every language against the PR build; every failing language re-run against the baseline build; each classified **clean / regression / pre-existing drift / broken**. Exit 0 / 1 / 2. Runs locally with the two existing venvs. Parallel (4 workers), so the sweep is faster than the old serial loop. |
| `tests/tools/tri_comparison_chart.py` | `--ci` fails closed when a baselined language has no data (that is the galaxyscope-off-PATH no-op, reproduced before the fix: 3 languages skipped, `all OK`, exit 0). Zero languages → 1. |
| `tests/tools/tree_sitter_accuracy_audit.py` | Zero languages → 1 (it already exited 1 on a missing galaxyscope; CLAUDE.md overstated that half). |
| `tests/tools/test_audit_runners_refuse_empty.py` (new) | 13 tests: classification, the exit-2 paths, label behaviour, step summary, both runners' zero-language refusal. |
| `docs/self_scan/ROSETTA_AUDIT.md` | Replaces `BUMPING_THE_ROSETTA_PIN.md`. The three outcomes, the label, local repro. |
| `CONTRIBUTING.md`, `docs/ecosystem.md`, `CLAUDE.md` | The three-step flow; the "two audits exit 0" paragraph rewritten to what is now true. |

### The flow after this

1. Engine PR merges first. `rosetta-audit` names the languages it moves; label it if intended.
2. Corpus re-bless PR against engine main; green by construction.
3. Nothing. keyword-rosetta's `bias-history.yml` regenerates the chart and keeps a "corpus owes a re-bless" issue in sync in between.

### Verified

- Local: `rosetta_audit.py` against the corpus with the main `.venv` — **46 checked, 0 regression, 0 pre-existing, 0 broken**. With galaxyscope off PATH / no corpus / zero folders → exit 2 each.
- `pytest` from an unrelated cwd: 21 passed (new file + `test_tri_comparison_chart_badge.py`). `audit_check.py`: all clear.
- This PR's own `rosetta-audit` run exercises the new workflow (the file is in its path filter). A follow-up throwaway commit will deliberately move one corpus count to show the **regression** classification firing, then be reverted.

### Cross-repo

- Companion: keyword-rosetta PR "ci: track engine main, retire ENGINE_REF, add bias-history" (link in a comment once open). Either can merge first; neither depends on the other's SHA any more.
- After both merge (admin): delete the `KEYWORD_ROSETTA_REF` variable (already unused), create the `rosetta:rebless-owed` label, note the retired pin check on #2669.

No engine code, manifests, ledger entries or golden masters change.


squid-protocol:
Cross-repo companion: squid-protocol/keyword-rosetta#46 (retires `ENGINE_REF`, adds `bias-history.yml`). Either can merge first.

squid-protocol:
**Regression-verdict demo, as promised.** Commit 30e989ac dropped `os.Open` from go's `io` rule (the corpus plants it in `main.go`), then was reverted in c66b62bd.

Run [33811217240](https://github.com/squid-protocol/gitgalaxy/actions/runs/33811217240) on the throwaway commit:

```
rosetta_audit: 1 failed here; re-running those against the baseline build .../baseline-venv/bin/galaxyscope
  REG  go                 FAIL go: 1 mismatch(es)
rosetta_audit: 46 language(s) checked -- 1 regression(s), 0 pre-existing, 0 broken.
```

Job failed with an annotation naming go; the 45 other languages stayed clean, so the verdict is per-language, not a whole-corpus red. Side effect worth noting: the 46-gate sweep took **85 s** (4 workers) against ~5 min for the old serial loop.

Issue #2729, opened 2026-09-04
arch_api conflates api-rule hits with converted orphans, and the rosetta gate reads the conflated column -- 36 of 44 languages pass api without the rule matching anything


squid-protocol:
Found by the #2669 F.4 census. This is the biggest of the three and it needs a decision before anything is changed.

## What the gate is actually asserting

The Contextual Baseline Fix (`galaxyscope.py` L2228) rewrites api in place when a file is imported:

```python
if popularity > 0 and orphans > 0:
    meta["equations"]["api"] = meta["equations"].get("api", 0) + orphans
    meta["equations"]["orphaned_logic"] = 0
```

`raw_arch_api` keeps the pre-fix value (#2536); `arch_api` carries the sum. keyword-rosetta's `verify_language.py` builds its column map from `RecordKeeper.SHORT_KEY_MAP`, which maps `api → arch_api` — the **conflated** column. Measured per file:

| file | manifest | api rule finds | orphans converted | gate reads |
|---|---:|---:|---:|---:|
| shell/a.sh | 3 | **0** | 3 | 3 ✓ |
| go/a.go | 3 | **0** | 3 | 3 ✓ |
| python/a.py | 6 | 3 | 3 | 6 ✓ |
| c/a.c | 12 | 9 | 3 | 12 ✓ |

Shell's api rule is `^export NAME` and no probe exports anything. Go's is `^func [A-Z]…` and every probe function is lowercase. Both measure **zero** and both pass, because three uncalled functions become three API surfaces. The manifests were recorded from the adjusted column, so the conversion is baked into the expectation.

Corpus-wide, `raw_arch_api` is 0 in 36 of the 44 languages that define an api rule. The median is therefore 0, and the bias report flags the ten languages whose rule *does* fire (c, python, embedded_python, fortran, lua, scala, livecode, css, makefile, html) as the deviants. The finding is inverted: those ten are the only ones measuring anything.

## Why it matters beyond the corpus

[...]

PR #2734, opened 2026-09-05
fix(core-engine): the api rule's own declarations are not new orphan surface (#2731)

### LINKED ISSUE #2731: arch_api counts a declared-public, uncalled function twice: 3 functions record api 6
Fell out of the corpus wave for #2729. It could not be seen before that wave, because no corpus language had both halves at once.

## The measurement

`galaxyscope.py` L2228 adds a file's `orphaned_logic` to its `api` when the file is imported. It does not ask whether those functions were *already* counted as API surface by the language's api rule. So a function that is both **declared public** and **uncalled** contributes twice.

keyword-rosetta `data/go/a.go`, three functions, all three exported (`func ProbeGlobals`, `func ProbeTest`, `func ProbeSafety`), none called in-file, the file imported by `main.go`:

| column | value |
|---|---|
| `function_count` | 3 |
| `raw_arch_api` (the api rule) | 3 |
| `raw_state_slop_orphans` | 3 |
| `arch_api` (what scoring reads) | **6** |
| `state_slop_orphans` | 0 |

Three functions, six units of public surface. The same three identifiers are counted once as declarations and once as converted orphans.

## Why it was invisible until now

Before the #2729 wave the corpus never planted an explicit public surface: `raw_arch_api` was 0 in 36 of 44 languages, so `arch_api` was *entirely* converted orphans and the two terms never overlapped. Now 30 languages declare their probes public and every one of them shows the doubling. Corpus-wide the effect is a bimodal `arch_api`: 6 where a language can declare visibility, 3 where it cannot but its orphans still convert, 0-1 where neither happens.

## Why it matters

`_calc_api_exposure` and `_calc_documentation` read the adjusted `api`. On real code the same overlap is the common case, not the exception: a library's public functions are exactly the ones with no in-repo caller. Every exported-but-not-internally-called function in every imported file is currently scored as two public surfaces, so `risk_api_exposure` and `risk_documentation` are inflated most for the codebases that are most library-shaped.

It also lands on the rosetta bias table as 21 unexplained cells (`risk_api_exposure` in 18 languages, `risk_documentation` in livecode/ruby/tcl) whose measured input `raw_arch_api` sits exactly on the median — the deviation enters through a column the corpus cannot see, which is the same structural gap #2729 fixed for the gate.

## Fix shape

Convert only the orphans the api rule did **not** already count. The recorder already has both numbers (`raw_arch_api`, `raw_state_slop_orphans`, #2536), so the cheap form is to credit `max(0, orphans − rule_hits)`; the precise form is to exclude, per function, the ones whose declaration the api rule matched — `spatial_maps` has the api rule's match locations and the function spans are known, which is the same span data #2727 needs.

Either way this is a golden-master change: it moves `arch_api` on every imported file in the real-world corpus, and `risk_api_exposure` / `risk_documentation` with it. Verify the direction on language-crucible before blessing.

Part of #2669. Siblings: #2729 (the gate now asserts the rule, done), #2727 (the orphan census is name recurrence, not a call graph), #2730 (the api rule has no stated contract).

squid-protocol:
Closes #2731. Part of #2669.

## The double count

`galaxyscope.py`'s Contextual Baseline Fix converts an imported file's orphans into API
exposure. It never asked whether the language's own `api` rule had already counted those
declarations, so a function that is **both declared public and uncalled** contributed twice.

keyword-rosetta `data/go/a.go` — three exported functions, none called in-file, imported by
`main.go`:

| column | before | after |
|---|---|---|
| `function_count` | 3 | 3 |
| `raw_arch_api` | 3 | 3 |
| `raw_state_slop_orphans` | 3 | 3 |
| `arch_api` (what scoring reads) | **6** | **3** |
| `risk_api_exposure` | 5.939 | 3.022 |
| `risk_documentation` | 79.155 | 40.613 |

## The fix

`detector.py` reports `api_declared_orphans` — how many of the file's orphans the `api` rule
matched **at their own declaration** — and the conversion credits only the remainder
(`orphans - api_declared_orphans`, clamped). The debt wipe stays unconditional: the file is
imported, so an already-public orphan is surface the rule counted already, not dead weight.

**The test is by name, not by span.** An orphan's name occurs exactly once in the whole file —
that is precisely what the orphan census just proved (`token_counts[name] <= 1`) — so a name
appearing on a line the `api` rule matched can only be its own declaration. No false positive is
possible. Span containment (`hit_vector["api"] > 0`) is both looser and tighter than that:

- **looser** — C's `api` rule matches any non-`static` declaration-shaped line, *including*
  body-local `int counter = 0;`, so every long function suppresses itself. Measured: the span
  form suppressed all 73 orphans in `c/cpython/ceval.c` and all 27 in `frameobject.c` (both
  files of `static` helpers); the name form suppresses 6 and 0.
- **tighter** — the public marker often sits outside the slicer's own span: JS/TS `export`
  precedes `start_idx`, php's span starts a line early, a java `@Test` line pulls `start_line`
  back off the `public` one. The span form scored javascript/typescript 0-of-3.

New `api_declared_orphans` is consumed by the orchestrator only — it is not a signal, does not
reach the recorder schema, and appears in no output artifact (checked against the audit JSON,
LLM digest, SBOM and the `file_data` schema).

## Verification

- **Golden masters re-blessed, both modes.** Every one of the 1254 full-precision / 1253
  zero-dependency differences is attributed, with nothing left over:

  | class | count |
  |---|---|
  | Topological Coordinates X/Y/Z (mass re-solve, corpus-wide) | 853 |
  | API Exposure % | 87 |
  | Documentation Exposure % | 71 |
  | `Exposed API / Public Exports` | 61 |
  | Structural Magnitude | 61 |
  | Directory Group Magnitude / `total_mass` | 29 + 40 |
  | avg api_exposure / documentation roll-ups | 50 |
  | ecosystem `composition/*/impact`, `health/avg_documentation` | 11 |
  | forensic top-3 rankings (`api_exposure/highest`, `undocumented_critical_path`) | 2 |

  61 files move on the real-world corpus, **all downward, zero increases**, and no structural
  signal other than `api` moves at all. A control A/B (same venv, `origin/main` vs this branch)
  measured total `arch_api` 35892 → 35127 (−2.13%).
- **keyword-rosetta:** 21 of 46 languages move — 20 go 6 → 3 (three probe functions, three units
  of public surface), `c` goes 12 → 9 / 10 → 7, `makefile` 8 → 7. The other 25 are unchanged
  (their `api` rule marks export lists or non-function surfaces, which never overlapped the
  orphan set). The corpus re-bless follows as a keyword-rosetta PR against this merged engine,
  per the no-pins flow.
- `pytest tests/` — 7568 passed (3 new regression tests: the go double count, the C
  static-vs-exported discrimination, and the orchestrator's credit/clamp arithmetic).
- `ruff_audit.py --ci`, `mypy_audit.py --ci`, `dead_key_audit.py --ci` — all clean against
  baseline. `tree_sitter_accuracy_audit.py` / `tri_comparison_chart.py` measure function/class/arg
  extraction and never read `api` (neither tool contains the string); this change cannot move them.

## Follow-up for the corpus side

#2729's gate pins `api_orphan_credit` (`arch_api - raw_arch_api`) as its only assertion that the
`main → a → b → c` import chain resolved. That credit legitimately becomes 0 for every language
whose probes are all declared public. `raw_state_slop_orphans 3` alongside `state_slop_orphans 0`
still proves the conversion ran, so the gate has a replacement available; it needs to be swapped
in the same PR that re-blesses the manifests.

### DIFF
--- a/gitgalaxy/core/detector.py
+++ b/gitgalaxy/core/detector.py
@@ -1328,6 +1328,7 @@ class StructuralExtractor:
 
             orphan_count = 0
             duplicate_count = 0
+            orphan_names: list[str] = []
             func_names = [f.get("name", "") for f in functions]
             func_name_counts = collections.Counter(func_names)
 
@@ -1383,9 +1384,39 @@ class StructuralExtractor:
                     elif len(func_name) > 3 and token_counts[func_name] <= 1:
                         # If the function name only exists where it was defined, it's an orphan
                         orphan_count += 1
+                        orphan_names.append(func_name)
                         usage_status = 1  # 1 = Orphan / Unused
 
                 func["usage_status"] = usage_status
+
+            # --- #2731: WHICH ORPHANS DID THE api RULE ALREADY COUNT? ---
+            # galaxyscope.py's Contextual Baseline Fix converts an imported
+            # file's orphans into API exposure. A function that is BOTH declared
+            # public AND uncalled was counted twice by that conversion -- once by
+            # the language's own `api` rule at its declaration, once as a
+            # converted orphan -- which is the common shape in library code,
+            # where the exported functions are exactly the ones with no in-repo
+            # caller. Count the overlap here (the orchestrator has no code text)
+            # and let the conversion credit only the remainder.
+            #
+            # The test is by NAME, not by span: an orphan's name occurs exactly
+            # once in the whole file (that is what the census above just proved),
+            # so a name appearing on a line the api rule matched can only be its
+            # own declaration -- no false positives are possible. Span
+            # containment would be both looser and tighter than that: looser
+            # because a long body can hold an unrelated api hit (C matches every
+            # non-static local declaration), tighter because the marker can sit
+            # outside the slicer's own span (JS/TS `export` precedes start_idx,
+            # php's span starts a line early, a java `@Test` line pulls start_line
+            # a line back off the `public` one).
+            api_declared_orphans = 0
+            if orphan_names and threat_locations.get("api"):
+                code_lines = code_stream.splitlines()
+                api_line_tokens: set[str] = set()
+                for line_no in set(threat_locations["api"]):
+                    if 0 < line_no <= len(code_lines):
+                        api_line_tokens.update(re.findall(r"\b\w+\b", code_lines[line_no - 1]))
+                api_declared_orphans = sum(1 for name in orphan_names if name in api_line_tokens)
 
             if orphan_count > 0:
                 equations["orphaned_logic"] = orphan_count
@@ -1429,6 +1460,10 @@ class StructuralExtractor:
                     round((file_token_mass / 1000000) * 3.00, 5) if file_token_mass is not None else None
                 ),
                 "threat_locations": threat_locations,
+                # #2731: how many of `orphaned_logic`'s functions the `api` rule
+                # already counted as public surface. Consumed by galaxyscope.py's
+                # Contextual Baseline Fix; never a signal in its own right.
+                "api_declared_orphans": api_declared_orphans,
             }
             if profile_regex:
                 result_payload["regex_telemetry"] = regex_telemetry
--- a/gitgalaxy/galaxyscope.py
+++ b/gitgalaxy/galaxyscope.py
@@ -2227,9 +2227,34 @@ class Orchestrator:
             if popularity > 0 and "equations" in meta:
                 orphans = meta["equations"].get("orphaned_logic", 0)
                 if orphans > 0:
+                    # #2731: credit only the orphans the language's own `api`
+                    # rule did NOT already count. A function that is both
+                    # declared public and uncalled used to contribute twice --
+                    # keyword-rosetta's `data/go/a.go` has three exported,
+                    # uncalled functions and recorded raw_arch_api 3 + orphans 3
+                    # = api 6, three functions for six units of public surface.
+                    # The overlap is the common case, not the exception: a
+                    # library's public functions are exactly the ones with no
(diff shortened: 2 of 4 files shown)


squid-protocol:
CI: everything green except `rosetta-audit`, which is the expected corpus-side red and is the fix's own confirmation.

All 21 failing languages report **only** `api_orphan_credit expected 3, got 0` (makefile: 1 file) —
64 mismatches, not one of them any other column, and the language list matches the local sweep
exactly. The credit is `arch_api - raw_arch_api`, so it going to 0 *is* the double count
disappearing: the api rule counted these declarations, and the orphan conversion no longer counts
them a second time.

Per the no-pins flow the corpus re-blesses against merged `main`, so this clears with the
keyword-rosetta PR that regenerates the 21 manifests. That PR also needs to replace what
`api_orphan_credit` was asserting for #2729 (that the `main → a → b → c` chain resolved) — the
credit is legitimately 0 now wherever every probe is declared public. `raw_state_slop_orphans 3`
with `state_slop_orphans 0` still proves the conversion ran.

`crucible-audit` (both modes), `tree-sitter-accuracy-audit`, `tri-comparison-audit`,
`ast-accuracy-audit`, `full-suite`, `ruff`/`mypy`/`dead-key` and all 12 smoke-test matrix cells
pass.

squid-protocol:
Labelled `rosetta:rebless-owed` — the documented mechanism for an intentional corpus-visible change. `rosetta-audit` now **passes**, reporting the same 21 languages as warnings rather than failures, so the PR is fully green. The re-bless is still owed on the corpus side exactly as described above.

Issue #2817, opened 2026-09-06
state_mutation: python, embedded_python and ruby never count a plain re-assignment (narrower than the #2765 contract)


squid-protocol:
Found by the `state_mutation` contract audit (#2765, `docs/state_mutation_rule_contract.md`).

The contract says one hit is *a statement that writes a new value into state that already exists*, and corollary 1's fallback says that where a language has no declaration syntax the assignment statement is the only write form there is, so it counts. shell, php, powershell, tcl, livecode, matlab, fortran, groovy and lua (for non-`local` names) all read that way.

python and embedded_python do not. Their rule counts `self.x = v`, `global`/`nonlocal`, `:=`, and the container mutators (`.append(`, `.update(`...), and never a plain `x = v`:

```
count = count + 1        # lua: 1, php: 1, shell: 1, go: 1 -- python: 0
self.count = count + 1   # python: 1
```

ruby is the same for locals (its rule counts `@ivar =`, `@@cvar =`, mutators and bang methods).

Both read 2 on the keyword-rosetta corpus because the plants are container writes, which *are* writes, so nothing is red -- but the same construct is scored in one language and invisible in the next, which is exactly the inequality #2812 exists to remove. The contract doc records python/ruby as **narrower than the contract, deferred**.

**Why it is not in the #2765 PR:** the fix is one arm per rule (a statement-start `x = v` with the kwargs/default-parameter shapes excluded -- black-formatted code puts spaces around a statement's `=` and none around a keyword argument's, which is the usable anchor), but python is the engine's most-scanned language and `risk_state_flux` for every python file in the telemetry population reprices. That deserves its own golden-master review, a crucible incidence table for the kwargs/multi-line-call false positives, and its own corpus re-bless, not a line inside a 30-language PR.

**To do:** the arm, strict tests for `x = 1` / `foo(x=1)` / `def f(x=1)` / `x == 1` / `x: int = 1`, the same for embedded_python and ruby locals, and a row update in `docs/state_mutation_rule_contract.md` moving the three from *deferred* to *agrees*.

Part of #2812 (Phase 3).

PR #2818, opened 2026-09-06
feat(contracts): state_mutation gets a stated contract — a statement that writes to existing state (#2765, Phase 3)

### LINKED ISSUE #2765: state_mutation has no stated contract: three incompatible definitions across 46 languages, and it governs the corpus's two worst metrics
Same pass as #2763/#2764 (2026-09-05, engine `3b3568e0`, corpus main `6183c4b`). This is the #2730 treatment applied to the next rule that needs it, chosen because the two metrics it governs are the corpus's **worst two by consistency score**:

| metric | consistency (share of languages in the ±25% band) |
|---|---|
| `risk_state_flux` | **42%** — worst of 49 |
| `state_mutation` | **60%** — second worst |

Eight languages sit red on `state_mutation` against a planted intent of 2 (assembly 6, c 5, go 5, dockerfile 4, haskell 4, m4 4, matlab 4, perl 4), and twenty sit red on `risk_state_flux`. Every one of those cells is ledgered, but the ledger entries describe *what each rule matched*, never whether the rules agree on what a mutation is. They do not.

## Three incompatible definitions, all live

Running each language's own rule over the corpus's identical planted intent:

**(a) Named mutation verbs** — only re-assignment counts.

```
m4       pushdef, m4_append, popdef, popdef
scheme   set!, vector-set!, ...
tcl      set|lappend|dict set|array set|incr|append
haskell  IORef, IORef, modifyIORef, writeIORef
```

**(b) Mutable-binding declarations** — declaring a variable counts, mutating it does not.

```
rust     \bmut\b|\.borrow_mut\(\)|Cell::|RefCell::|Atomic[A-Za-z0-9]+
kotlin   \b(var|MutableList|MutableMap|...)\b
swift    \b(var|inout|mutating|didSet|willSet)\b
zig      \bvar\b|\.\*[ \t]*=[^=]
scala    \b(var|scala\.collection\.mutable|Atomic...)\b
```

**(c) The assignment operator** — every `=` counts, including initialization.

```
c        (?<![=!<>])=(?![=])|\*(?!\s*const)\w+[ \t]*=|(?:\+\+|--)
go       :=|(?<![=!<>])=(?![=])|<-|\bappend\(|\batomic\.(?:Add|Store|Swap)
lua      \b[a-zA-Z_]\w*(?:\[[^\]]+\]|\.[a-zA-Z_]\w*)?\s*(?<![=<>~])=(?![=])
groovy   ^[ \t]*\w+(?:\.\w+)*[ \t]*=(?!=)
fortran  (?!\b(?:KIND|LEN|UNIT|FMT|FILE|STATUS|ACTION)\s*=)\b[A-Za-z_][...]{0,199}[ \t]*=[^=>]
```

These are not three renderings of one idea. A `let x = 5` is invisible to (a), invisible to (b) if the binding is immutable, and a mutation to (c). A `var x: Int` with no write is a mutation to (b) and invisible to (a) and (c). The corpus plants exactly two mutations per language and reads 2 to 6 depending only on which family the language landed in.

## Two outright false positives found while classifying

**haskell counts a type signature.** `a.hs` records `state_mutation` 2, and both hits are the word `IORef` in a *declaration*:

```haskell
counter :: IORef Int          -- hit
region  :: IORef Int          -- hit
```

The rule's alternation `\b(IORef|STRef|TVar|MVar|TMVar|...)\b` matches the type name anywhere. A type annotation naming a mutable cell is not a mutation — `modifyIORef`/`writeIORef` in the same alternation are, and `b.hs` reads those correctly. haskell's `state_mutation` is therefore 4 against a plant of 2, and `risk_state_flux` 15.5 against a median of 7.751 (+100%), from two type signatures.

**dockerfile double-counts its own `globals` plants.** `a.dockerfile` records `state_mutation` 2, both from `ENV`:

```dockerfile
ENV REGION=us-east
ENV HOME_ZONE=1
```

`ENV` is dockerfile's `globals` rule *and* the first arm of its `state_mutation` rule (`^[ \t]*ENV[ \t]+[a-zA-Z0-9_]+`). Declaring a build-time environment variable is a global declaration; it is a mutation only under family (c)'s reading, which dockerfile otherwise does not use (its second arm is `export X=`). Every Dockerfile in every scan pays `state_mutation` once per `ENV` line.

## What a contract would have to say

`docs/api_rule_contract.md` (#2730/#2743) is the model: one sentence, then corollaries, then a 46-language audit table. The sentence for this rule is a real design choice, not a lookup — the two candidates are

* **"a write to state that outlives the expression"** — family (a) and (c)-minus-initialization. Declarations never count; `mut`/`var` count only at the write site. This is what `risk_state_flux` reads as if it were true.
* **"the introduction or use of mutable state"** — family (a), (b) and (c). Declarations count. This is what rust/kotlin/swift/scala/zig implement, and it makes the signal a *style* measure rather than a behaviour measure.

Either is defensible; today's mixture is not, and the mixture is what the ×3 cascading-flux amplifier (`spatial_correlation.py:278`) then multiplies. Whichever is chosen, the corollaries the audit needs:

1. A type annotation naming a mutable container is not a mutation (haskell, and check scala's `Atomic*`/`mutable` arms and kotlin's `Mutable*` for the same shape).
2. A token already owned by another rule for the same construct is not a second signal (dockerfile `ENV` ↔ `globals`; check c/go/lua/groovy/fortran `=` against their own `globals` plants for the initialization case).
3. Initialization is a declaration, not a mutation, under contract 1 — which is the c/go/lua/groovy/fortran audit and the largest bucket.

## Method, and why it is cheap

`tools/screen_plant.py` and `_registry.py` in keyword-rosetta run any candidate through the live compiled rules; `language-crucible/data/<lang>` gives the real-world incidence per language before anything is changed, the way #2743's audit table was built (`c` 8675 → 1141 was found exactly this way). Expect the same shape of result: a few languages widen, most narrow, and the golden-master diff is attributable per language.

## Corpus pairing

`state_mutation` is planted (2 per language, SPEC.md's probe table), so every language's manifest re-blesses off one scan. Ledger entries that narrow or retire: `state-flux-branch-weighting`, `batch4-dual-keyword-overlaps` (dockerfile/matlab/m4 halves), `probe-idiom-collides-with-measured-signal`, and the `state_mutation` arm of `string-literal-selective-shielding`.

Part of #2669. Sibling of #2730 (api) — same shape, next rule.

squid-protocol:
Phase 3 of #2812, first family. `docs/state_mutation_rule_contract.md` is the contract, the corollaries, the fallback family and the 46-language audit; `gitgalaxy/standards/signal_contracts.py` row `state_mutation` goes `draft -> stated` (3 of 68).

## The contract

> **`state_mutation` matches a statement that writes a new value into state that already exists: a re-assignment (plain, compound or `++`), an in-place update of a container or structure, or a write through a mutable cell or reference.**

The issue's two candidate sentences were a real choice. The rejected one ("the introduction or use of mutable state") is what rust/kotlin/swift/scala/zig and js/ts implemented and it makes the signal a style measure; every consumer (`_calc_state_flux`, cognitive load, safety, the cascading-flux pair, the `cascading_state_mutation` bottleneck) reads the count as behaviour, so the count is behaviour.

Corollaries, each found live: (1) a declaration is not a write, even `let mut`/`var`/`int x = 5`/go `:=` -- and where a language has no declaration syntax the assignment statement is the write; (2) a type or modifier naming mutable state is not a write (`IORef Int`, `AtomicInteger x;`, `volatile`, `mutable`); (3) a read, bind or cast is not a write (`this.` rvalues -- 376 hits of `this.error(` in one TS file -- haskell `<-`, perl bare `shift`, cpp `&`, `std::move`, `payable()`); (4) one statement is one hit, and a token another rule owns is not a second signal (dockerfile `ENV` -> globals; matlab `clear`, m4 `popdef`, apex `.clear(` -> cleanup; `UPDATE t SET` once; `s = append(s, x)` once).

## What changed

- **30 rules** rewritten on one shared shape: a statement-start + bare-lvalue anchor (a type or `let`/`var` in front breaks the match) with a trailing-comma guard for enum members / named arguments, `++`/`--` touching an operand, and receiver-anchored container mutators. Per-language reasoning is in each rule's `# #2765 contract:` comment and the doc's audit table.
- **Real-world incidence** (language-crucible v1.2.0, code stream): c 10644 -> 6300, cpp 11623 -> 4005, rust 2331 -> 636, typescript 9069 -> 5618, perl 5838 -> 3932, lua 6025 -> 2248, sqlite 35 -> 4, haskell 41 -> 1. dart 764 -> 2052 and abap 48 -> 402 *widen*: plain assignment had been invisible in both (dart's old arm only matched unspaced `x=1`; abap's rule had no `=` at all).
- **Corpus:** 21 languages' `state_mutation` cells move (`rosetta_audit.py`: 21 regressions, all attributed). Rule-side: haskell a.hs 2->0, dockerfile a 2->0, c a 2->0 / b 3->2, go b 5->2, objective-c 3->2, perl 3->2 & c 1->0, m4 c 2->0, apex c 1->0, matlab c 1->0. Plant-side (authored to the old rules, re-planted as real writes in the corpus PR): cpp, java, javascript, kotlin, rust, scala, swift, typescript, zig, solidity, sqlite, abap; livecode/scheme/tcl carried a third write in the decoy line. The re-blessed corpus verifies **46/46** against this branch.
- **Golden masters** blessed from a clean `language-crucible` v1.2.0 worktree (a second bless from the dirty sibling checkout matched byte for byte). `tests/tools/bless_scope.py` on the diff: 9,363 per master = 4,323 topological + 5,040 substantive, whose leaf keys are exactly the state signal, its weighted view, `Amplified Cascading Flux`, and the formulas that read it (State Flux / Cognitive Load / Error & Exception exposure, Structural Magnitude, the race-condition pair). One artifact newly parses: `cobol/.../RESPSTR.cpy` lost ~500 phantom `END-STRING` hits and no longer trips the aperture's "Packed Payload Guard".
- **Tests:** 11 per-language strict pairs flipped; the three "intentional double-classification" tests the contract retires (dockerfile ENV, m4 popdef, matlab clear) rewritten to assert the contract; one cross-language module `tests/extraction/languages/test_state_mutation_contract_2765.py` (positives/negatives per language, one-statement-one-hit counts, ReDoS detonation for every rewritten regex). Full suite 7,895 passed; ruff/mypy/dead-key audits clean; `signal_contract_audit.py --ci` exits 0 with the baseline shrunk by one.

## Tooling (the process ask)

Every rule session so far rebuilt the same two scripts; they are now standing tools: `tests/tools/rule_probe.py` (a rule's incidence and matched lines on both corpora, `--json` snapshots, `--compare` emits the audit-table rows) and `tests/tools/bless_scope.py` (the uncapped golden diff bucketed by section / language / leaf key, plus artifacts that moved between parsed and excluded). The `rule-contract-audit` skill is rewritten as a command recipe with the parallel legs and the worktree/env pitfalls this session hit.

## Deferred, with an issue

python, embedded_python and ruby never count a plain local re-assignment (`x = v`): narrower than the contract by construction, not red on the corpus, and a population-wide reprice of `risk_state_flux` for the most-scanned language -- filed as #2817 rather than folded in here.

## Cross-repo

Engine merges first. keyword-rosetta branch `rebless/gitgalaxy-2765-state-mutation` (PR to follow) carries the re-plants, 20 manifest cells, the notes, a new ledger entry `state-mutation-contract-2765` and the retirements in `batch4-dual-keyword-overlaps`; its `verify.yml` goes green once main carries this PR. `bias-history.yml` regenerates the chart on merge.

Closes #2765


squid-protocol:
Cross-repo: the corpus re-bless is squid-protocol/keyword-rosetta#79 (branch `rebless/gitgalaxy-2765-state-mutation`, verified 46/46 against this branch, bias report regenerated at full precision: `state_mutation` 2 in all 43 comparable languages, `risk_state_flux` open-defect share 0/45, corpus open-defect share 79 → 62 of 2303). This PR merges first; #79 goes green once main carries it.

Issue #2827, opened 2026-09-06
api_declared_orphans tokenizes api lines with `\b\w+\b`, so a hyphenated or qualified name is never recognised as already-declared (#2731's double count returns for cobol and scheme)


squid-protocol:
`detector.py`'s `api_declared_orphans` (#2731, the input to `galaxyscope.py`'s orphan → api
conversion) decides whether the `api` rule already counted an unreferenced function by tokenizing
each api-matched **line** with `\b\w+\b` and testing the orphan's name for membership:

```python
api_line_tokens.update(re.findall(r"\b\w+\b", code_lines[line_no - 1]))
api_declared_orphans = sum(1 for name in orphan_names if name in api_line_tokens)
```

`\b\w+\b` cannot produce a token containing `-`, `:`, `.`, `!` or `?`. Every function whose name
holds one is therefore invisible to the test, the overlap reads 0, and the conversion credits the
orphan a second time — which is exactly the double count #2731/#2734 was built to remove.

This is the same tokenizer family as **#2754**, where the census itself counted with `\b\w+\b`
and every hyphenated Scheme define, `Verb-Noun` PowerShell function, `::ns::proc` and
`Class::method` read as an orphan regardless of use. #2754 fixed the census and left this
consumer on the old tokenizer.

## Measured

Per-file over `keyword-rosetta/data`, engine `fix/2823-export-list-declarations`, comparing the
current `\b\w+\b` line tokenizer against `_name_boundary_pattern(name)` searched over the same
api-line text:

| language | `api_declared_orphans` today | with a name-aware test | why |
| --- | --- | --- | --- |
| `cobol` | 0 | 9 | `PROBE-GLOBALS`, `PROBE-TEST`, … — hyphenated paragraph names, 3 per file in a/b/c. **Present on main today** |
| `scheme` | 0 | 9 | `(export probe-globals)` — 3 per file in a/b/c |
| every other language (42) | 292 | 292 | unchanged; all-`\w` names |

`haskell` is the control that shows the mechanism working correctly when the names cooperate: its
export list is `module A (probeGlobals, probeTest, probeSafety) where`, all three names are
plain `\w`, `api_declared_orphans` reads 3, and `api_orphan_credit` correctly stays 0.

## Consequence

`api_orphan_credit` (= `arch_api − raw_arch_api`) over-reads for these languages, and it feeds
[...]

Issue #2806, opened 2026-09-06
orphaned_logic reports 100% dead code for five languages whose invocation model never names the callee — the family #2727 closed without covering


squid-protocol:
`orphaned_logic` reports **every function in the file as unused** for five languages whose invocation model does not name the callee. #2727 fixed a different half of this metric and its closing comment says explicitly that it did not reach this one — but the issue is closed COMPLETED and the ledger entry still carries `disposition: upstream-bug`, `still_reproduces: true`, pointing at it.

Measured 2026-09-06 on engine main `a59ecf6b` / corpus main `2038a86`. Ledger: `orphan-detection-is-name-recurrence`.

## The measurement

Baseline is 2.50 orphans/file — 13 probe functions across 4 files, 3 of them called from main. Five languages sit at 3.25, which is 13/4: *every* function orphaned.

```
raw_state_slop_orphans, per file, corpus median 2.50

3.25  abap, jcl, m4, makefile, objective-c      +30%, amber -- every function reads unused
2.75  cobol                                     +10%, in band (see below)
2.50  median
0.25  scheme                                    -90%
```

cobol is the near-miss that proves the mechanism rather than an exception to it: 11 of its 13 paragraphs read orphaned, and the two that clear are `DISPATCH-PARA` and `PROBE-BRANCH` — the only two paragraph names that appear a second time anywhere in the corpus, in `main.cbl`'s `ALTER DISPATCH-PARA TO PROCEED TO PROBE-BRANCH.` The nine copybook paragraphs, entered through `ENTRY 'PROBE-X'` alternate entry points, are named once each and all read dead.

## What #2727 fixed, and what it said it did not

#2727 replaced a whole-file token-frequency test with a span-scoped name-reference test. `_is_orphan` (`detector.py` ~L7086) now counts occurrences of the name outside `[start_idx, end_idx)`, with a declaration discount when the span contains none, and #2777/#2774 have since corrected the boundary pattern and discounted `_visibility_export` names. That work is sound and this issue does not touch it.

Its closing comment, verbatim:

[...]

Issue #2774, opened 2026-09-05
orphan census: an `export -f name` line counts as a use, so a declared-public uncalled function can never be reported dead


squid-protocol:
Same pass as #2769–#2773 (2026-09-05, engine `3b3568e0`, corpus main `6183c4b`). `raw_state_slop_orphans` is the corpus's third-worst metric at 61%. #2754 fixed the census's *reference test*; #2768 covers the length guard in front of it. This is the third piece, and it is the one with real-world consequences: **naming a function in an export statement clears its orphan flag.**

## Repro

`keyword-rosetta/data/shell`, whose `a.sh` defines three probes and calls none of them. Two scans, identical but for the three `export -f` lines:

```
                      raw_state_slop_orphans
with    export -f            a.sh 0   b.sh 0   c.sh 0
without export -f            a.sh 3   b.sh 3   c.sh 3
```

`export -f probe_globals` is `a.sh`'s `api` plant — the only way a shell script publishes a *function* rather than a variable (#2730 widened the rule to reach it). It also names `probe_globals` on a line outside the function's span, which is exactly what `_is_orphan` counts:

```python
outside = len(word.findall(code_stream[:start_idx])) + len(word.findall(code_stream[end_idx:]))
if inside == 0:
    outside -= 1          # the declaration itself
return outside <= 0       # detector.py:6995
```

One export line, one outside occurrence, flag cleared. The function is still never called.

## Scope

Five languages in the corpus reach this shape because their `api` idiom names the function: assembly, powershell, ruby, shell, tcl — all sitting at `raw_state_slop_orphans` 0.25 against a 2.50 median (−90%). The ledger entry `api-export-names-suppress-orphan-credit` records the trade and its own text calls the outcome out: *"a mention in an export statement is outside the span, so it will still clear the flag — the honest fix is a real call graph."*

Outside the corpus it is worse, because the affected shape is the common one in library code:

* `export -f name` (bash), `Export-ModuleMember -Function name` (PowerShell), `module_function :name` / `public :name` (Ruby), `namespace export name` (Tcl), `__all__ = ["name"]` (Python), `export { name }` (JS/TS), `pub use module::name` (Rust), a C header's prototype.
* In every one of those, the export list is the file's public API — and the functions on it are precisely the ones with no in-repo caller, which is the definition of what the orphan census is looking for.

[...]

PR #2828, opened 2026-09-06
fix(core-engine): an export construct that names many functions is still a declaration (#2823, roadmap Phase 3)

### LINKED ISSUE #2823: unreferenced_by_name reads 0.25/file for haskell and scheme: a module export list and a type signature are declarations, not references
`unreferenced_by_name` (the census #2806 renamed from `orphaned_logic`) reads **0.25 per file for haskell and scheme against a 2.50 corpus median** — 12 of each language's 13 probe functions read as referenced when nothing calls any of them. This is #2806's defect with the sign flipped: there the census counted a function as unreferenced that the language reaches without naming it; here it counts a function as referenced on a name occurrence that is a *declaration*, not a reference.

Measured 2026-09-06 on engine `fix/2806-orphan-invocation-model` (branched from main `87cfc933`) / corpus main `5cf94e7`, per-file census over `keyword-rosetta/data`.

## The two forms

`_is_orphan` (`detector.py`) counts occurrences of the name outside `[start_idx, end_idx)`, discounts **one** of them when the span contains none (the declaration, which for several slicer modes falls outside its own span), and discounts the offsets captured by the registry's `_visibility_export` rule (#2774). That budget of one implicit + one declared form is exactly enough for tcl/shell/ruby/powershell/assembly. haskell and scheme each write the name a third time:

```haskell
module A (probeGlobals, probeTest, probeSafety) where   -- export list: one occurrence per name
probeGlobals :: Int -> Int                              -- type signature: another
probeGlobals env = env                                  -- the equation (span starts after the name)
```

```scheme
(export probe-globals)          ;; scheme declares no _visibility_export, so this reads as a use
(define (probe-globals env) env)
```

## What is and is not enough

- **scheme** needs only the `_visibility_export` declaration it never got. Adding `^\([ \t]*export[ \t]+([a-zA-Z][\w\-!?*/<>=+]*)` moves the corpus 0.25 → **3.00** per file, not 2.50 — because `main.scm`'s `entry` does not call its probes either, so scheme also carries the corpus plant gap #2806 found in abap/m4/makefile/objective-c, hidden until now behind the export false-negative. Both halves have to land together or scheme swaps an amber cell for a worse one.
- **haskell** cannot be expressed with the mechanism as it stands. `export_name_starts` is built as `frozenset(m.start(1) for m in _vis_export.finditer(...))` — one capture group, one name per match — and a Haskell module export list holds an arbitrary number of names inside one pair of parens. Declaring the type signature instead (`^([a-z]\w*)[ \t]*::`) is a legal single-capture rule but does **not** move the count (measured: still 0.25), because the export list alone is enough to clear every flag.

So the fix is (1) let `_visibility_export` yield more than one name per match — every group, or a second `_visibility_export_list` whose capture is the whole list and whose names are split by the engine — and (2) declare it for haskell and scheme, and (3) plant the missing dispatch calls in `keyword-rosetta/data/{haskell,scheme}/main.*` in the same wave.

## Why it matters

The metric feeds `risk_tech_debt` and the orphan→api conversion, so an under-reading language looks maximally clean on both. It is also the last shape of the census that is still measurement rather than code: after #2806, `orphan-detection-is-name-recurrence`'s remaining languages are these two plus the no-functions family (html, markdown, yaml, sqlite, dockerfile, css), which is a different question (#2549's undefined family).

Part of #2812. Successor to #2774 and the sign-flip of #2806; related #2727, #2754.

squid-protocol:
Closes #2823. Part of #2812.

`unreferenced_by_name` read **0.25 per keyword-rosetta file for haskell and scheme against a 2.50 corpus median** — 12 of each language's 13 probe functions reading as *referenced* while nothing called any of them. That is #2806's defect with the sign flipped: there the census called a function unreferenced that the language reaches without naming it; here it called a function referenced on an occurrence that is a **declaration**, which corollary 2 of the stated contract says must not clear the flag.

## One outside occurrence each, and it is the export

```haskell
module A (probeGlobals, probeTest, probeSafety) where   -- outside the span
probeGlobals :: Int -> Int                              -- INSIDE the span
probeGlobals env = env                                  -- INSIDE the span
```

```scheme
(export probe-globals)                                  -- outside the span
(define (probe-globals env) env)
```

The Haskell type signature needed **no rule of its own**: haskell's slicer span *starts* at the signature, so that repetition is already inside the function's own definition. That explains the issue's own unexplained measurement — declaring `^([a-z]\w*)[ \t]*::` as an export form moves the count not at all.

## `_visibility_export_list`

#2774's `_visibility_export` captures **one** name per match, which is the whole statement for `export -f foo` and its four siblings and cannot express a construct that names every exported function at once. This adds the plural form: its capture groups are **regions** holding exported names, and `_export_declaration_offsets` records the start offset of every name token inside a region.

Two keys rather than one generalised key, deliberately. The singular form records its group's own start offset **verbatim**, and assembly exports `.foo` while ruby exports `save!` — whose first and last characters a region tokenizer does not treat as part of a name. Tokenizing those captures would put the offset one character off and silently stop discounting them. A language declares one key or the other, and `test_the_two_export_keys_never_disagree_about_a_language` is the gate on that.

haskell declares the module header's parenthesised list (bounded by the `) where` anchor, since `[^)]*` stops at the first `Opt(..)`, and every one of pandoc's seven headers terminates exactly right); scheme declares both of its clause forms, R7RS `(export a b c)` and Guile `#:export (a b c)`.

## Measured, both corpora

| corpus | before | after |
| --- | --- | --- |
| keyword-rosetta, haskell | 0.25 /file | **3.00** /file |
| keyword-rosetta, scheme | 0.25 /file | **3.00** /file |
| keyword-rosetta, other 44 languages | — | **unchanged** |
| language-crucible, haskell | 2 of 147 units (1.4%) | **41** of 147 (27.9%) |
| language-crucible, scheme | 0 | **0** — neither file uses an export clause outside a macro template |

No other language *can* move: the singular path is byte-identical and only these two declare the new key.

**3.00, not the 2.50 median**, because both languages also carry the same plant gap #2806 found in abap/m4/makefile/objective-c — hidden until now behind the export false-negative — where `main`'s `entry` dispatched one of its three probes. The corpus PR plants the other two and both land on 2.50.

**The crucible's 39 new hits are correct.** Every one is a pandoc utility whose name occurs exactly three times in its own file — export list, type signature, equation — with no caller there. That is corollary 5 working as designed: `Text.Pandoc.Shared` is a library module of 100 exported helpers, 36 of which nothing in that file calls, and saying so is the input `galaxyscope.py`'s orphan → api conversion needs. Same shape as #2774's shell precedent (0 → 3 per file).

## The bless

**Nine value differences, identical in both fixtures, every one of them `haskell/pandoc`:**

| what | was | is |
| --- | --- | --- |
| `Shared.hs` Unreferenced By Name | 0 | 36 |
| `Filter.hs` Unreferenced By Name | 2 | 4 |
| `Options.hs` Unreferenced By Name | 0 | 1 |
| those three files' Tech Debt Exposure | 0.0% / 41.35% / 0.0% | 97.04% / 85.83% / 9.7% |
| `haskell/pandoc` avg tech debt | 3.76 | 17.51 |
| global `avg_tech_debt` | 27.332 | 27.386 |

The tech-debt rows are the census's own weight arriving downstream (`slop_stress` counts each unit at 2.0). **No topological `X`/`Y`/`Z` coordinates moved at all** — unusual for a bless, and the reason is that no node's structural mass changed.

## One neighbour moved — a pre-existing defect this made visible

With scheme finally producing orphans, its `api_orphan_credit` went 0 → 3 in a/b/c: the double count #2731/#2734 exists to prevent. `api_declared_orphans` tokenises the api-matched line with `\b\w+\b`, which cannot produce a token holding a hyphen, so `probe-globals` is never recognised in `(export probe-globals)`.

**haskell is the control that proves the diagnosis** — its probe names are all-`\w`, its `api_declared_orphans` reads 3, and its credit correctly stays 0.

Not new, and much larger than the control corpus suggests: cobol carries it on main today, and over the crucible `api_declared_orphans` moves 4,983 → 5,122 under a name-aware test, +139 across zig (`@"..."` quoted identifiers), lua (`bit.bor`, `M.start`) and powershell — whose count is **0 for the entire language**, because a `Verb-Noun` name always contains a hyphen. Filed as **#2827** rather than fixed here: it moves five languages and needs its own bless.

## Docs

- `docs/unreferenced_by_name_contract.md`: corollary 2 gains the plural form, the audit table's haskell/scheme row moves 0.25 → 2.50, and a new section states the whole finding — why two keys, what the crucible says, and what the bless moved.
- `how_to_add_a_language.md` rule 19 gains the **opposite-mistake** bullet: a census that reads too *clean* is the same defect as one that reads too dirty.

## Verification

- Full suite: **7,914 passed**, 4 skipped, 9 xfailed.
- `crucible_check.py`: both fixtures re-blessed, diff scoped above (9 values each, no unattributed mover).
- `rosetta_audit.py --baseline-bin <engine main>`: 46 languages, **0 regressions**.
- `tri_comparison_chart.py --all --ci`: 3 languages, all OK.
- `audit_check.py` (ruff / mypy / dead-key / ast-accuracy): all clear.

## Cross-repo

- Corpus: **squid-protocol/keyword-rosetta#83** (branch `plant/2823-haskell-scheme-dispatch`) — plants the two dispatch calls, re-blesses scheme's `api_orphan_credit` cell against a new #2827 ledger entry, and narrows `orphan-detection-is-name-recurrence` from 8 languages to 6.
- **This merges first**; that one goes green against gitgalaxy `main` once it lands.
- Follow-on: #2827 returns scheme's and cobol's `api_orphan_credit` cells to 0 and owes another corpus re-bless.


squid-protocol:
## `rosetta-audit` failure: the companion re-bless, labelled

```
REG  scheme             FAIL scheme: 3 mismatch(es)
       - a.scm: api_orphan_credit expected 0, got 3
       - b.scm: api_orphan_credit expected 0, got 3
       - c.scm: api_orphan_credit expected 0, got 3
rosetta_audit: 46 language(s) checked -- 1 regression(s), 0 pre-existing, 0 broken.
```

This is the exact mirror of squid-protocol/keyword-rosetta#83's own red CI, and the two together are the whole cross-repo state: `rosetta-audit` checks out the corpus at `main`, whose scheme manifest still expects `api_orphan_credit: 0` — the value it reads when `(export probe-globals)` is still wrongly clearing the orphan flag. With this build there are orphans again, `galaxyscope.py`'s Contextual Baseline Fix converts them, and the cell reads 3.

The corpus PR blesses those three cells to 3 against a new ledger entry (`api-declared-orphans-blind-to-non-word-names`, #2827) and cannot go green until this merges; this cannot go green until that one does. Applied **`rosetta:rebless-owed`** — the label the workflow documents for exactly this shape.

Worth stating explicitly: **only scheme moves, and only on this one downstream cell.** haskell passes here, on corpus main, with no plant — its `api_orphan_credit` is 0 in both engine states, for two different reasons (no orphans before; all three suppressed by `api_declared_orphans` after). The census itself is not gated in any manifest, so haskell's and scheme's 0.25 → 3.00 move is invisible to this audit; what it caught is the single consequence two layers downstream, which is #2827.

Issue #109, opened 2026-07-05
# [REFACTOR] Upgrade `SignalProcessor` Risk Equations for Class-Level Aggregation


squid-protocol:
## 🎯 Objective
Modify the core execution loop in `signal_processor.py` to aggregate Function-level risks into their parent Classes before normalizing them into File-level scores.

## 🛠️ Implementation
* `prism.py` already successfully links functions to classes (`func["parent_class_name"] = cls["name"]`). 
* Update the `_calc_*` methods (e.g., `_calc_algorithmic_dos`, `_calc_cognitive_load`) to iterate through `classes` first.
* Sum the function impacts within the class boundary, generate a Class-level risk score, and then pass those Class aggregates up to the File-level Sigmoid normalizer.

squid-protocol:
Dug into the actual code to ground this before starting. Two corrections and one reframing:

**Correction: the class-linkage isn't in `prism.py`.** It's `detector.py:503` (`func["parent_class_name"] = cls["name"]`), built as part of the class extractor at `detector.py` ~lines 466-522. That same block already computes `method_count`, `state_entanglement`, and `lcom_score` per class (confirmed, matches what #110 describes as the current `class_data` state) -- so the class-detection infrastructure is further along than this issue implies, it's just misattributed.

**Reframing: don't touch the `_calc_*` sigmoid methods.** I read `_calc_cog_load` closely (the pattern holds across the other 13). These are density/sigmoid formulas -- they divide by the *whole file's* LOC and apply file-scale normalization (sigmoid slope/offset, doc-coverage cooling, a Gini multiplier). But per `08-01-methodology.md`'s table, **Level 2 (Class) is "count based, 0-∞"** -- the same category as Function level. Only **Level 3 (File) is "Sigmoid normalized, 0-100."**

So "iterate through classes first, generate a Class-level risk score, pass those aggregates up to the File-level Sigmoid normalizer" isn't quite the right shape -- sigmoid normalization shouldn't run at the class level at all, and File-level output shouldn't change numerically (summing via classes vs. summing directly from all functions is mathematically identical, since class membership partitions the function list exhaustively).

[...]

PR #2903, opened 2026-09-08
makefile dead_code: a `# Label:` prose comment is not a commented-out target (#2851)

### LINKED ISSUE #2851: makefile dead_code reads any `# Word:` prose comment as a commented-out target
## Summary

`makefile`'s `dead_code` rule reads **any `# Word:` prose comment as a commented-out target**, because it is the only comment-anchored `dead_code` rule in the registry that is *shape*-anchored rather than keyword-anchored.

```
makefile  ^[ \t]*#[ \t]*(?:[a-zA-Z0-9_./%+-]+[ \t]*::?|(?:[a-zA-Z0-9_.-]|\+(?!=))+[ \t]*(?::|\?|::)?=|\b(?:ifeq|ifneq|ifdef|ifndef|include)\b)
shell     #[ \t]*(?:if|for|while|function|export|echo|printf|cd|rm|sudo|ls)\b
python    #[ \t]*(?:def|class|import|if|for|while|try|return)\b
yaml      ^[ \t]*#[ \t]*(?:-?[ \t]*run:|uses:|jobs:|steps:|script:)
```

The other three anchor on a **closed set of real keywords**. makefile's first alternative is `<identifier><optional space>:` — which every `Label:` prose comment satisfies. `# TODO:`, `# NOTE:`, `# HACK:`, `# author:`, `# Copyright:` are all read as commented-out rules.

## Evidence (keyword-rosetta control corpus, engine `9b4889e`)

makefile records `dead_code = 7`, and **every one of the 7 is prose**:

| matched comment | count | what it actually is |
|---|---|---|
| `# decoy:` | 4 | the corpus's own shell furniture, one per file |
| `# HACK:` | 1 | the planted `fragile_debt` probe |
| `# TODO:` | 1 | the planted `planned_debt` probe |
| `# author:` | 1 | the planted `ownership` probe |

Not one is a commented-out makefile construct. Two of the seven are worse than noise: they **double-count the debt plants**, so `dead_code` moves with `fragile_debt`/`planned_debt` for reasons unrelated to dead code.

For contrast, the same corpus's shell and python files carry the identical `# TODO:`/`# HACK:` comments and correctly record `dead_code = 0` — their rules require a real keyword.

## Impact

- `dead_code` feeds `risk_dead_code` (`_registry.risk_dependencies`), so a language-wide false positive propagates into a published risk score.
- It fires on the single most common comment convention there is (`Label: text`), so real-world Makefiles will carry it broadly — this is not a corpus artifact.
- It is the only `dead_code` rule that can match a comment containing no language construct at all.

## Fix shape

Anchor the target alternative on something a prose label cannot satisfy. A commented-out rule is nearly always followed by its recipe or by real target syntax, so the cheapest correct narrowing is to require what a target line actually has and a prose label does not — e.g. a dependency list or a recipe continuation after the colon (`# foo: bar`, `# foo:` + tab-indented line), or at minimum reject a colon that ends the comment or is followed only by prose words. The variable-assignment and `ifeq|ifneq|...` alternatives are already keyword- or operator-anchored and are not implicated.

## References
- Found by the keyword-rosetta non-planted-rule sweep (scoring the "non-planted keyword extraction" group), which classified 37 nonzero cells across 7 signals; this was the only one that was not language idiom or an overlap with a deliberately planted token.
- Sibling rules that get it right: `shell`, `python`, `yaml` `dead_code`.

squid-protocol:
## What this does

makefile's `dead_code` rule read **every `# Label: text` prose comment as a commented-out target**, because its target alternative was `<identifier>[ \t]*::?` -- a bare shape, where every sibling comment-stream rule (shell, python, yaml) anchors on a closed keyword set. Make's target syntax has no keyword, so the rule now demands evidence a prose label cannot supply.

Closes #2851

## How bad it was

The issue measured the control corpus: all 7 of makefile's `dead_code` hits were prose (`# decoy:` x4, `# HACK:`, `# TODO:`, `# author:`), two of them double-counting the `fragile_debt` / `planned_debt` plants, and `dead_code` feeds `risk_dead_code`.

The golden-master crucible could not see it (its one Makefile carries no such comment: 0 before, 0 after). So the rule was measured before it was touched on a **3,002-file deduplicated real-world Makefile pool** (one Linux kernel headers tree, dpkg/python/arduino/git `.mk` files) -- the kind of corpus with rich comments:

| | hits | files with a hit |
|---|---:|---:|
| old rule | 3,360 | 2,807 / 3,002 |
| new rule | 283 | 57 / 3,002 |

**2,753 of the 3,360 old hits were `# SPDX-License-Identifier: GPL-2.0`** -- every kernel Makefile. The rest: `# Compile: create object files...`, `# Author:`, `# Note:`, `# FIXME:`, `# https://...`, and option tables (`#  -Wall...: warning level`). Every real commented-out target in the pool survives the narrowing: `#build: lib`, `# all: $(PRG).elf lst text eeprom`, `# clean :`, `#install: install_lib ...`, `# %.html: %.txt`, `# text/%.text: %.xml %.txt`, `# comedi.h.xml : $(INC_UAPI)/linux/comedi.h`, `#$(LUFA_PATH)/LUFA/LUFA_Events.lst:`, `#install-html: html`, `# quick-install-man:` (+ recipe).

## The rule

A commented target counts when:

- **(a)** its name is a lifecycle target the `api` rule already names (`all|install|build|clean|test|run`) plus the GNU standard-target set (`check|dist|distclean|mostlyclean|maintainer-clean|uninstall|installcheck|installdirs|install-{strip,html,dvi,pdf,ps}`; `info|ps|html|pdf|dvi` left out -- they are also prose labels);
- **(b)** its name is a file / pattern / variable form (`# %.o: %.c`, `# lib.a: $(OBJS)`, `# $(TARGET): $(OBJS)`, `# .PHONY: x`) -- never a leading `-` (option tables) or a trailing `.` (`# e.g.:`);
- **(c)** the next comment line is a commented recipe (`#\t...`), the shape a rule takes when it is commented out whole.

In every form the prerequisite list must read as Make prerequisites (tokens, `$(...)` refs, `|`, an optional `;` recipe), so a `,`, a `:` (URLs) or a sentence period ends the claim; a multi-target rule carries at most three co-targets (`# build parallelizes well and finishes roughly at once:` is a sentence); form (c) wants a single lowercase / file-form name on a space-indented line (`# vmlinuz is:` + a tab-indented listing, a tab-continued `# NOTE:\t...` paragraph and a pasted `#\tavrdude: ...` log are all prose). Two same-class exclusions on the sibling alternatives: a commented assignment cannot start with `-` (`# -mcmodel=medium breaks modules`), and `include` (now also `-include`/`sinclude`) needs a path-like argument (`# include the stub's deps` is prose).

**Residue, named not hidden.** A plain-name target with identifier prerequisites and no recipe (`# foo: bar`) is ambiguous with prose (`# note: see`) and is deliberately not claimed. A commented recipe line on its own (`#\t$(CC) ...`) is not counted either: `#\t` also opens tab-aligned prose. The assignment alternative still reads a `key = description` prose table (`# c89 = "ANSI" C`, ~40 of the 283 remaining pool hits) -- not this issue's shape, the issue called that alternative "not implicated", and it is an order of magnitude smaller than the target defect.

## Also in this PR

- `tests/tools/rule_probe.py` now honours a language's `exact_matches` names. It filtered on `extensions` only, so it skipped every crucible `Makefile` and reported the corpus as `--` for a language the golden master does scan.
- `how_to_harden_extraction.md` bug class 45: a comment-stream rule anchored on a *shape* rather than a keyword reads prose labels; the code-stream twin (`func_start`) can be shape-only legitimately, and a small hand-picked corpus can carry none of the `# Label:` habit while every kernel Makefile does.

## Verification

- `test_makefile_strict.py`: 32 positive shapes, 36 negative shapes (the 7 control-corpus lines, the SPDX line, the real-world prose shapes), a one-rule-one-hit count, a "debt / ownership plants are not double-counted" check, and 8 ReDoS payloads (scaling measured at ~8x per 8x input, 2k -> 16k chars, for every quantified piece). 203 pass.
- `rule_probe.py dead_code makefile`: rosetta **7 -> 0** (4 files), crucible **0 -> 0** (1 file).
- `ruff_audit --ci`, `mypy_audit --ci`, `dead_key_audit --ci`, `audit_check.py`, `signal_contract_audit.py --ci`: all clear.
- Full `pytest tests/` (minus the golden crucible, which is the bless below): **8681 passed**, 4 skipped, 9 xfailed, 3 xpassed.
- `rosetta_audit.py` against the companion corpus branch: **46 languages checked, 0 regressions, 0 pre-existing, 0 broken**.
- `crucible_check.py`: **4 differences on each fixture, 0 topological**, all on `c/cpython/Makefile.pre.in` (`dead_code` 14 -> 4 and its derived Commented Logic Exposure). The ten dropped lines are prose (`# FIXME:`, `# bpo-21536:`, `# BEGIN: freezing modules`, `# mode: makefile`, `# End:`); the four kept are commented-out `MODULE_EGG_*` / `PROFILE_TASK` assignments. Nothing newly parsed or excluded. Re-blessed.

## Cross-repo

Corpus PR: squid-protocol/keyword-rosetta#109 (ledger: `makefile-dead-code-reads-prose-labels` retired with the verifying scan; the `non-planted-rules-match-idiom-and-planted-tokens` NOT-COVERED clause annotated; bias report regenerated at full precision). No manifest cell moves: `dead_code` is unplanted. Engine merges first.


squid-protocol:
Windows matrix red on the first push: `ERROR at setup of test_makefile_dead_code_2851_redos_immunity[...]` on all four Python legs, green everywhere else. Not the regex -- pytest exports each node id as `PYTEST_CURRENT_TEST`, and a 50,000-character ReDoS payload inside a parametrize id exceeds Windows' 32,767-character cap on one environment variable (`ValueError: the environment variable is longer than 32767 characters`). Fixed in e1935d0a by keying the payloads with short `ids=` (longest node id now 126 chars); the trap is recorded in the `ci-push-checklist` skill so the next payload table does not repeat it.

Issue #3034, opened 2026-09-14
Native graph engine: shared CSR graph index + parity/benchmark harness


squid-protocol:
Part of #3033.

## Scope

Build the integer-indexed CSR index once from `_resolve_edges()`; add a test-only parity harness (`networkx` as oracle, seeded random graphs plus the crucible graph) and a benchmark script.

Move the existing `_pagerank` to the index. Keep its float-summation order unless a scoped golden diff shows no movement.

## Acceptance criteria

- The index is used by PageRank
- The parity harness exists
- Native PageRank beats `nx.pagerank`'s warm 4.3 ms on the crucible graph (it is 10.2 ms today)
- Both golden masters are unchanged
- This one must land first

## Relevant baseline

| operation | networkx | notes |
|---|---|---|
| pagerank | 4.3 ms (scipy) | native `_pagerank` today: **10.2 ms** (dict-of-str). Must beat nx once on integer arrays. |

## Networkx conventions to match

- PageRank uses scipy's implementation via `nx.pagerank` when scipy is available
- Native port must match the Pagerank algorithm's output exactly

squid-protocol:
**Revised by the tailoring decisions on #3033 (2026-09-14).**

The parity harness must support two oracle modes:
- **strict:** networkx's current call, for metrics whose definition is unchanged: SCC, articulation points, assortativity, Louvain
- **tailored:** networkx computing the NEW definition, for example
  - `nx.betweenness_centrality(G)`: exact, unweighted
  - per-source `nx.single_source_shortest_path_length` for directed reachable-pair path length
  - uncapped `nx.descendants` / `nx.ancestors`

Acceptance is unchanged otherwise. Also add a work-budget helper (deterministic, e.g. V×(V+E) ≤ threshold, never wall-clock) that every metric uses instead of node-count cutoffs.

Issue #3033, opened 2026-09-14
Native graph engine: replace networkx with a stdlib-only implementation (zero-dependency, exact parity, faster)


squid-protocol:
# Native graph engine: replace networkx with a stdlib-only implementation

Supersedes #3031. Builds on #3027 / PR #3032.

## Why

- **#3027 / PR #3032:** made PageRank native (`_pagerank()` in `gitgalaxy/core/network_risk_sensor.py`) and runs it in BOTH modes, so there is no mode drift and no networkx-version drift. Every metric still not computed is `None`, never `0.0`.
- **networkx declares no dependencies of its own**, but `nx.pagerank` (3.x dispatches to `_pagerank_scipy`) and `nx.degree_assortativity_coefficient` import numpy/scipy. Verified by blocking those imports: pagerank and assortativity fail; betweenness, closeness, Louvain, modularity, SCC, avg path length, articulation points and descendants still work. So `pip install networkx` alone gives a scan that believes it is full precision but silently loses metrics.
- **Once networkx leaves the runtime,** graph metrics are identical on every install. Zero-dependency mode then loses nothing on the graph side; only tiktoken, ML (xgboost/pandas/numpy) and pyyaml remain optional (see `docs/zero_dependency_mode.md`).

## What networkx still does in the engine (the inventory)

`gitgalaxy/core/network_risk_sensor.py` (`build_dependency_graph`):
- `nx.betweenness_centrality(G, k=min(V,100) if V>500 else None, weight="weight", seed=42 if sampled else None)`
- `nx.closeness_centrality(G)`, bypassed (`None`) when V > 1500
- `community.louvain_communities(U, seed=42)` (fallback `greedy_modularity_communities`), then `community.modularity(U, communities)`, skipped when V > 5000. `U = G.to_undirected()`.
- `nx.degree_assortativity_coefficient(G)`
- `nx.strongly_connected_components(G)`, giving cyclic density = nodes in SCCs of size > 1, divided by V
- `nx.average_shortest_path_length(U.subgraph(largest connected component))`, skipped when V > 5000
- `nx.articulation_points(U)`, count

`gitgalaxy/security/security_auditor.py` (`_resolve_dependency_graph`):
- `min(len(nx.descendants(G, path)), 500)` and `min(len(nx.ancestors(G, path)), 500)` per file.
- A pure-Python BFS fallback (`get_nth_degree`) already exists. It differs from nx in two edge cases:
[...]

Issue #3027, opened 2026-09-14
Zero-dependency mode: networkx metrics that were never computed read as 0.0 outside the DB (false 'Containment' insight, 0.0 macro table, HITL never fires)


squid-protocol:
## Symptom

In zero-dependency mode, networkx-derived metrics that were **never computed** read as `0` / `0.0` in every output except the SQLite DB. There they are NULL (`record_keeper.py`, `if session_meta.get("zero_dependency_mode")`). A 0.0 PageRank or blast radius looks like a real measurement of a leaf file. Several consumers then draw conclusions from it:

| Where | What happens without networkx |
|---|---|
| `network_risk_sensor._fallback_build_graph` | Writes `pagerank_score` / `normalized_blast_radius` / `betweenness_score` / `closeness_score` = `0.0` and `systemic_threat_vector = []`. |
| `signal_processor.py` AI topology (~L1252-1281) | Sorts AI files by an all-zero pagerank, then appends **"Containment (Low Risk): The AI components are safely isolated at the edge of the network…"** because `pr < 0.2`. That is a false claim. |
| `llm_recorder.py` §3.5 macro table (~L375-395) | `if net_macro:` is truthy (the keys exist with `None` values), and each cell renders `or 0.0`. So the brief prints Modularity 0.0, Cyclic Density 0.0%, Articulation Pts 0. This undoes #473's None convention one hop downstream, which is exactly what `network_risk_sensor.py`'s #473 comment warns against. |
| `llm_recorder.py` per-file topology (~L1039-1057) | Blast radius, betweenness and closeness print as 0.0. |
| `signal_processor.py` `systemic_bottlenecks` (~L2154-2203) | All scores are 0, so brief §13.8 is dropped with no note saying why. |
| `dev_agent_firewall.py` L42/L67 | `requires_hitl` is keyed on blast radius, so it can never fire. The DB column `requires_hitl` reads 0. |
| `archetype_classifier.py` L71-72, L108/L117 | `log_pagerank`, `log_blast_radius` and `pagerank_gini` go into nearest-centroid as zeros, so the file/repo composition archetype can differ from full precision. The module docstring acknowledges this. |
| `supply_chain_firewall.py` ~L232-238 | The blast-radius/betweenness multiplier stays 1.0. |

(`--max-systemic-threat` had the same problem. It now warns and is skipped without networkx, as of the #3024 PR.)

## Ask

[...]

PR #3044, opened 2026-09-14
Native graph engine: closeness + average path length, native in both modes (#3037)

### LINKED ISSUE #3037: Native graph engine: closeness at every size + directed average path length
Part of #3033.

## Scope

Closeness on the reversed index with Wasserman-Faust; average path length over the largest undirected component.

Replace the V>1500 (closeness) and V>5000 (avg path length) node-count cutoffs with a work budget on V×E. The crucible graph (2,817 nodes) computes exact closeness in 21 ms with networkx, yet is skipped today.

Parallel per-source BFS above a size threshold, with results combined in fixed order.

## Acceptance criteria

- Parity with networkx wherever networkx computes the value
- Average path length beats 1,213 ms on the crucible graph
- Note: raising the cutoff CHANGES full-precision output (closeness starts being computed on large repos). That needs a scoped bless and a PR note.

## Relevant baseline

| operation | networkx | notes |
|---|---|---|
| closeness (exact, all nodes) | 21.4 ms | the 1,500-node cutoff is far too conservative for sparse import graphs |
| avg shortest path length (LCC) | **1,213.5 ms** | the dominant cost of the whole graph phase |

## Networkx conventions to match

- **closeness** (directed): computed on `G.reverse()`, i.e. incoming distance
- Unweighted BFS (`single_source_shortest_path_length`), `totsp = sum(dist)`
- Value is `(len(sp)-1)/totsp` when `totsp>0` and `V>1`
- `wf_improved=True` multiplies by `(len(sp)-1)/(V-1)`
- **avg path length:** unweighted, largest connected component of U, mean over all ordered pairs

squid-protocol:
Closes #3037. Part of #3033 (the native graph engine). Builds on #3042 (#3034, merged): it uses that PR's `GraphIndex` and `WorkBudget`.

## What changed

`graph_engine.closeness_and_path_length(index, budget)` runs one breadth-first search per node over the reversed CSR index. That single pass produces both hop-count metrics:

- **Closeness** (per file): strict parity with `nx.closeness_centrality(G)`. That is directed, incoming distance, in Wasserman-Faust form. Registered in the parity harness as `strict`.
- **Average path length** (repo): **redefined** per the tailoring decisions on #3033.
  - **Now:** the mean hop count over every ordered pair (A, B) where A transitively imports B, across all files. It is `None` when no file imports another.
  - **Before:** the mean shortest path in the largest *undirected* component. That ignored import direction and covered 22.5% of the crucible's files.
  - It is registered in the harness as `tailored`. The oracle is per-source `nx.single_source_shortest_path_length`.
- **Cutoffs → a work budget.** The V>1500 (closeness) and V>5000 (path length) node-count cutoffs are gone. `PATH_METRICS_WORK_BUDGET` is 50M incoming-edge scans, about 3 s of pure-Python search in the worst case. It counts work, never time, so the golden masters can't flake. Past the budget, both metrics are `None`, never `0.0`.
- **Both modes.** Like PageRank (#3027), this runs in the networkx and zero-dependency builders alike, from the same index. Zero-dependency mode therefore gains closeness, the "fragile dependency chain" ranking, and path length.

Edge weight plays no part in either metric: it stays a PageRank strength. networkx's weighted paths would treat it as a distance, which is backwards (#3033).

## Measured on language-crucible v1.2.0 (V=2,817, E=1,354; warm, best of 5, `tests/tools/graph_parity.py --db`)

| metric | parity | native | before |
|---|---|---|---|
| closeness | strict, max \|diff\| 0.0 | 1.72 ms | 12.87 ms (nx), and **skipped** at this size |
| avg path length | tailored, max \|diff\| 0.0 | 1.61 ms | about 1.2 s (nx, undirected LCC) |

Both metrics come out of one pass: 1.7 ms in total.

## ⚠️ Definition change: temporal-crucible / snapshot history

`repo_data.network_avg_path_length` keeps its column, but its meaning changes with this PR. Values from before and after are **not comparable**. On language-crucible it was 7.0723 and is now 1.358. `docs/zero_dependency_mode.md` records this under "Older snapshots", and the brief's §3.5 interpretation text now describes the new definition. Closeness was NULL in zero-dependency mode and above 1,500 files before this PR, so older snapshots show NULL where newer ones have values.

## Golden masters: re-blessed after an uncapped, scoped diff

Scoped with `golden_diff.deep_compare` over a fresh scan of each mode, using `GITGALAXY_DISABLE_GIT_HISTORY=1` and the same flags as `test_golden_crucible.py`.

- **full precision: 2 differences, both intended**
  - `network_macro/avg_path_length`: 7.0723 → 1.358 (the redefinition)
  - `systemic_bottlenecks/fragile_dependency_chain`: `[]` → the top closeness × error-risk files. Closeness used to be skipped above 1,500 files, which covers the whole crucible.
- **zero dependency: 2 differences, both intended**
  - `avg_path_length`: `None` → 1.358
  - the same `fragile_dependency_chain` ranking as full precision
- Closeness is identical in the two modes for all 2,817 files: 0 mismatches between the two scans' DBs. Before this PR, all 2,817 were NULL in both modes; now 300 are non-zero.
- No coordinate or other metric moved in either mode. Closeness feeds no mass or risk formula.
- The only commit on main since the fixtures were last blessed is #3042, which moved neither fixture. So every difference above belongs to this PR.

## Verification

- The parity harness passes on 25 seeded random graphs for each metric, and on the real crucible graph (CLI above).
- New sensor tests:
  - a 1,501-file chain in both modes: closeness and path length are computed (formerly skipped) and equal the closed forms. The long chain also shows the search is iterative.
  - the work budget exhausted → `None`, not `0.0`
  - a graph with no imports → path length `None`
- Updated tests: zero-dependency closeness equals full precision. A networkx centrality failure or macro-block failure no longer takes down closeness or path length, since both are computed outside those blocks.
- `pytest` passes on the 9 affected suites (308 passed), run as plain `pytest` from outside the repo.
- `audit_check.py`: all clear.


squid-protocol:
Closing: every change in this PR is already on main. #3046 was stacked on this branch, so its squash merge (492b0292) carried #3037's commits too, including the f63649d9 lambda cleanup. Nothing is left to merge here.

Issue #2773, opened 2026-09-05
args counts call-site arguments in objective-c (+62%) and typescript: a coupling signal that scales with how many calls a file makes


squid-protocol:
Same pass as #2769–#2772 (2026-09-05, engine `3b3568e0`, corpus main `6183c4b`). `args` sits at 73% consistency and `avg_func_args` at 80%. Most of the low tail is the documented no-parameter-surface family (css, html, dockerfile, makefile, sqlite, yacc, cobol, and yaml until #2753). The **high** tail is a different thing and has never been examined: two languages count the arguments at *call* sites as well as the parameters at declarations.

## Measured

`galaxyscope --db-only` over `keyword-rosetta/data/objective-c` and `data/typescript`, `file_data.struct_args` against `struct_func_start`:

| file | `func_start` | `args` | the extra hits |
|---|---|---|---|
| `objective-c/a.mm` | 3 | **5** | `XCTAssert(kit)`, `NSAssert(value)` |
| `objective-c/b.mm` | 3 | **5** | `os_log(msg)`, `DDLogInfo(msg)` |
| `objective-c/c.mm` | 3 | **5** | `free(conn)`, `release(conn)` |
| `objective-c/main.mm` | 4 | **6** | `abort()`, `exit(payload)` |
| **objective-c total** | 13 | **21** (+62%) | 8 call sites |
| `typescript/a.ts` | 3 | **5** | `describe(kit)`, `expect(kit)` |
| **typescript total** | 13 | **15** (+15%) | 2 call sites |

Every extra hit is a *call*, and every one of those calls is a probe the SPEC planted for a different signal — `XCTAssert`/`describe`/`expect` are `test`, `os_log`/`DDLogInfo` are `telemetry`, `free`/`release` are `cleanup`, `abort`/`exit` are `high_risk_execution`. So `args` is not measuring a parameter surface for these two languages; it is re-counting other signals' plants, and a language's `args` total scales with how many function calls its shell happens to contain.

The two rules' relevant arms:

```python
# objective-c — the second alternative matches any `name: value` selector-shaped text,
# the third matches a bare call's parentheses
[...]

Issue #2732, opened 2026-09-05
jcl + yaml: dead_code and spec_exposure comment-stream rules missing where the comment stream exists (#2610 leftover)


squid-protocol:
Found by the epic #2560 n/a sweep of non-planted risk-formula inputs (child of #2669; na_check.py flags `jcl/dead_code`, `jcl/spec_exposure`, `yaml/spec_exposure` as unreviewed absences).

## What's missing

`dead_code` and `spec_exposure` are **comment-stream rules** — `detector.comment_analysis()` applies them alongside `doc`/`ownership`/`planned_debt`/`fragile_debt`. Among the 46 rosetta-covered languages, only three lack `spec_exposure`: markdown (deliberate — the generic `[SPEC-…]`/`[spec …]` bracket-tag pattern collides with markdown link syntax, now ledgered corpus-side), **jcl**, and **yaml**. jcl also lacks `dead_code`.

- **jcl**: both absences are the same shape #2610 fixed for the debt rules — before `_strip_jcl_comments`, jcl's comment stream was structurally empty, so no comment rule could ever fire; #2610/#2611 routed `//*` lines to `comment_analysis` and added `planned_debt`/`fragile_debt`/`ownership`, but `dead_code` and `spec_exposure` were never revisited.
- **yaml**: `#` comments have always reached the comment stream (yaml `comment_lines` measures nonzero), and because `spec_exposure` never sees the code stream, YAML's `[a, b]` flow-sequence syntax cannot FP against it — the usual reason to hesitate doesn't apply. The absence looks arbitrary: css, yacc, m4, solidity, html all define it.

## Evidence the morphology is real (jcl dead_code)

Commenting out a statement by turning `//` into `//*` is the universal JCL practice. The pool corpus (gitgalaxy/data) has **14 licensed .jcl/.prc files** carrying `//*`-commented-out statements, including whole JOB cards:

```
//*CREL005 JOB ,,CLASS=A,MSGCLASS=H,
//*CREL001 JOB ,,CLASS=A,MSGCLASS=H,
```

## Fix shapes

```python
# jcl — commented-out statements (comment stream = raw //* lines post-#2610).
# Keyword must be followed by operand-shaped text, not prose, to keep
# "//* PROC statements are..." banners out (observed FP shape in the pool):
"dead_code": re.compile(
[...]

Issue #2516, opened 2026-08-30
Epic: Legacy Mainframe Modernization


squid-protocol:
## Overview

Tracking epic for GitGalaxy's mainframe legacy-modernization initiative: hardening the engine's existing COBOL/JCL structural-signature extraction (especially CICS coverage) and adding first-class support for the rest of the mainframe language family needed for full CICS-platform visibility (BMS, HLASM, REXX, PL/I, DB2 SQL/SQL PL).

No epic previously tracked this as one initiative -- the `legacy-modernization` label existed but child issues were scattered and inconsistently tagged. This gives the initiative one home going forward.

## Language additions (new mainframe-family parsers)
- #2502 -- PL/I (`pli`), for full CICS visibility
- #2503 -- HLASM / IBM Z Assembly (`hlasm`)
- #2504 -- REXX (`rexx`)
- #2505 -- Divorce `.bms` from the `jcl` profile into its own BMS Map parser
- #2511 -- IBM DB2 SQL / SQL PL (`db2_sql`)

**Note:** #1142 (PL/I signatures, parented under the aerospace epic #75) overlaps with #2502 -- both add a `pli` profile with different signature-key emphasis. Cross-referenced on both issues; needs a human decision on consolidation before either is picked up, so they don't ship as two independent, conflicting implementations.

## CICS/JCL extraction hardening (existing cobol/jcl profiles)
- #2482 -- jcl args: continuation lines
- #2483 -- jcl args: multi-line over-counting
- #2484 -- cobol CICS channels and containers
- #2485 -- cobol CICS advanced queuing and exceptions
- #2486 -- jcl CICS utility execution and resources

## Related, already-resolved groundwork
Prior extraction-hardening and bug-fix work that shaped the current cobol/jcl baseline this epic builds on: #259, #292, #501, #590, #776, #850, #854, #1142, #1858, #1890, #1891, #1892, #1949, #1975, #2480. See the `legacy-modernization` label for the current full set.

## Scope note
Not every cobol/jcl-adjacent bug belongs here -- structural-extraction bugs shared across many languages (Mode A/B slicing, ReDoS bounding, etc.) that happen to touch cobol/jcl incidentally are tracked under the general extraction-hardening epic (#813) instead, even when a fix ships alongside this work.

squid-protocol:
**#2990 (COBOL/JCL semantic-coverage assessment) shipped as PR #3006.**

[...]

Issue #1898, opened 2026-08-20
prism.py: ABAP's positional comment stripper reuses Fortran/COBOL's column-1 anchor set, erasing real code as comments


squid-protocol:
## Summary

`prism.py`'s `_strip_positional_comments()` treats any line whose column-1 (or column-7) character is in the shared `POSITIONAL_ANCHORS` set as a full-line comment and strips it from `code_stream` entirely. That set — `{"*", "C", "c", "/", "!"}` (`gitgalaxy/standards/language_standards.py:215`, consumed at `gitgalaxy/core/prism.py:192`) — is built for Fortran's column-1 `C`-means-comment convention and COBOL's fixed-form markers, but is applied unconditionally to ABAP too, even though ABAP's own real column-1 comment marker is `*` only (confirmed by ABAP's own `lexical_family` doc comment in `language_standards.py`: *"the engine must monitor Column 1 for an asterisk `*` to identify line-level Commented / Non-Executable Text"*).

Since real ABAP class declarations are conventionally written flush-left (`CLASS <name> DEFINITION` / `CLASS <name> IMPLEMENTATION.`), column 1 is a literal `C` — so this shared anchor check silently erases every real class header into the comment stream.

## Isolated repro

```python
from gitgalaxy.core.prism import Prism
from gitgalaxy.standards.language_standards import LENS_CONFIG, LANGUAGE_DEFINITIONS

p = Prism(LENS_CONFIG.get("COMMENT_DEFINITIONS", {}), LANGUAGE_DEFINITIONS)
sample = "CLASS zcl_foo DEFINITION\n  PUBLIC\n  CREATE PUBLIC .\nENDCLASS.\n"
result = p.split_streams(sample, "abap")
print(result["code_stream"])
# '\n  PUBLIC\n  CREATE PUBLIC .\nENDCLASS.\n'  -- the CLASS header line is GONE
print(result["comment_stream"])
# 'CLASS zcl_foo DEFINITION'  -- misclassified as a comment
```

## Real-world impact

[...]

PR #3074, opened 2026-09-16
Add BMS as a language: divorce .bms from jcl into a dedicated 3270 map profile (#2505)

### LINKED ISSUE #2505: Language Reassignment: Divorce .bms from JCL into dedicated BMS Map parser
**Parent:** #2516

**Language**: BMS (Basic Mapping Support)
**Extensions**: `.bms`, `.map`
**Lexical Family**: `positional_anchored` (It is technically HLASM macro code).

### Scoping Situation
Currently, `.bms` is assigned as an extension under the `jcl` profile in `language_standards.py`. This is structurally incorrect, as BMS defines 3270 UI screens via Assembler macros (like `DFHMSD`, `DFHMDI`, `DFHMDF`), whereas JCL is a batch orchestration language.

We need code that covers:
*   **x (Extraction)**: Remove `.bms` from the `jcl` `extensions` list.
*   **y (Reassignment)**: Either create a dedicated `bms` language profile or subsume `.bms` into the new `hlasm` profile (since BMS maps are just HLASM macros). Given GitGalaxy's universal key mapping, a dedicated `bms` profile where macros map purely to `ui_framework` might yield the cleanest analytical signal for frontend discovery.
*   **z (Domain Sensors)**: `ui_framework` must capture the screen definitions (`DFHMSD`, `DFHMDI`).

### Implementation Requirements
1.  Update `language_standards.py` to divorce `.bms` from `jcl`.
2.  If creating a standalone profile, follow `how_to_add_a_language.md` and generate `tests/extraction/languages/test_bms_strict.py`.

squid-protocol:
Closes #2505 (parent epic #2516).

## What

`.bms` used to be an extension of the **jcl** profile, which was structurally wrong — a BMS map defines a CICS 3270 UI screen via HLASM macros (`DFHMSD`/`DFHMDI`/`DFHMDF`), while JCL is batch orchestration — so every map scanned as signal-free job control. This PR gives BMS its own dedicated profile, per the issue's x/y/z breakdown:

- **x (Extraction):** `.bms` removed from jcl's `extensions`.
- **y (Reassignment):** dedicated `bms` profile (`languages/bms.py`), the issue's recommended option — a mapset's macros map to `ui_framework` for the cleanest frontend-discovery signal. `hlasm` (#2516) doesn't exist yet, and a standalone profile keeps that door open.
- **z (Domain sensors):** `ui_framework` fires once per `DFHMSD`/`DFHMDI`/`DFHMDF` statement, anchored to HLASM operation-field position (a macro name in a comment banner or an `INITIAL='...'` literal never counts; continuations never double-count).

## Profile shape

- `func_start` = named map (`DFHMDI`), `class_start` = mapset header (`DFHMSD`, `TYPE=FINAL` excluded even when the closer repeats the name), `api` = both (the names `SEND MAP`/`MAPSET` reach from other members), `args` = **named** `DFHMDF` fields — the generated symbolic-map copybook's data surface (#2773 fallback family; unnamed fields are screen literals and declare nothing).
- HLASM `COPY` → `import` + `_dependency_capture`; conditional assembly (`AIF`/`AGO`/`SET*`/`MACRO`…) → `macros`; `*`-comment-anchored `dead_code` (operand-guarded, jcl's #2732 reasoning), `ownership`, debt and `[SPEC-n]` rules.
- `invocation_model: "positional"` (top-level, jcl's #2806 shape): nothing inside a BMS source reaches a map by its extracted name — the `POSITIONAL_LANGUAGES` pin in the #2806 contract test is widened deliberately, with the corollary-4 justification recorded.
- Everything else is explicit `None` per Strict Feature Parity, with stated-absence comments; strictness row `None` (declarative markup, blp's family) and `bms` joins the mainframe inert-matter untestable set.

## The `.map` collision

The issue names `.map` as an extension, but in the wild `.map` is overwhelmingly JS source maps and linker maps. Handled with existing mechanisms rather than a bare extension claim:

- `.map` added to `COLLISION_FREQUENCIES` (never Tier-1 locks on extension alone);
- `internal_discriminator` (the objective-c/matlab `.m` mechanism) locks a real BMS map named `.map` at Tier 2;
- `disqualifiers` (`.js`/`.css`/`package.json`/`.o`/`.elf`…, #377's toxic-neighbor mechanism) collapse the single-candidate ecosystem-gravity fallback, which would otherwise self-support `.map` with its own count and lock a source map to bms sight unseen (caught in end-to-end verification, not by unit tests — the gravity path only runs in a real scan).

## Prism

`_strip_positional_comments` gains a `bms_mode`: column-1 `*` and `.*` are the only comment markers, no column-7 check, no inline split. The shared `positional_anchored` anchor set (`{'*','/','C','c','!'}`) would have erased any statement whose column-1 name starts with one of them (`CUSTMAP DFHMDI` — the #1898 ABAP class-header bug, verbatim), and HLASM has no inline comment marker, so `!`/`*>`/`"` inside operands must never split.

## Verification

- `tests/extraction/languages/test_bms_strict.py` — 48 tests per the how-to's adversarial framework: per-signature positive/negative, schema completeness, the jcl-divorce pin, Prism family sanity (both comment styles stripped, anchor-char names survive, debt rules fire on the comment stream), exact counts on a realistic mapset, Rule-15 exclusion checks, re.M audit, scaled ReDoS payloads, and `.map` collision resolution through the real `LanguageDetector`.
- Full default suite green (the 10 `test_security_auditor.py` failures pre-exist on an untouched main checkout — environment, not this PR).
- `audit_check.py`: ruff/mypy/dead-key baselines, format, ast-accuracy all clear.
- **Differential scan:** `crucible_check.py` passes both fixtures with zero drift. Per Step 5.4 this is an *empty-corpus* zero — the crucible contains no `.bms`/`.map` files — not a verified-no-change zero; the strict suite and a scratch-repo end-to-end scan (`galaxyscope --db-only`: `.bms` → `bms`, sourcemaps never `bms`) carry the verification instead.

## Cross-repo

None required to merge. Follow-up owed: the **keyword-rosetta control folder** for `bms` (same follow-up state as pli, #3057).


squid-protocol:
Cross-repo update: the keyword-rosetta control folder this PR's description listed as a follow-up is now authored — companion PR squid-protocol/keyword-rosetta#133 (draft until this PR merges; that repo's CI checks out engine main). Verified against this branch: `verify_language.py bms` PASS, 100 assertions across 4 files, every count exactly as planted; a `verify.yml` dispatch with `engine_ref=pull/3074/head` is running there as pre-merge CI evidence. Merge order: this PR first, then #133.

PR #3196, opened 2026-09-19
feat(archetypes): retrain function/file/repo brains on the v2.9.0 corpus

squid-protocol:
Retrains all archetype models on the fresh **739-repo v2.9.0 fleet scan** (8.27 M functions, 1.49 M files) and bakes the new brains into the engine.

## What changed
| Model | Result |
|---|---|
| `GENERAL_FUNCTION_INFERENCE_MODEL` | k=14, re-labeled. 12 archetypes kept 1:1 (nearest-centroid on the aligned contract); **"Tests & Verification"** merged into **"Annotated & Test Methods"**; new **"Stateful Encapsulated Methods"** for the Encapsulated-Accessors split. Contract dropped `log_density_token_mass` (Rosetta governance) → 37-dim. |
| `GENERAL_FILE_INFERENCE_MODEL` | k=18, **83-dim** (now carries the 14 function-stoichiometry features from `func_cluster_ratios`). |
| `file_archetype_brain.json` | k=15, refrozen, **100% self-consistent**. |
| `repo_archetype_brain.json` | k=7, refrozen, **100% self-consistent**. `feature_contract_sha` re-pinned (composition features follow the retrained file archetypes). |

`DNA_SOURCES` (the schema-stable feature→hit_vector mapping the trainer doesn't emit) is carried over from the prior brains, verified to cover every new feature.

## Validation
- **Parity scan** (original_doom, baked engine): no dimension mismatch, **no `Unclassified` flood** — func 748/748, file 146/146, composition 146/146; new labels flow through ("Stateful Encapsulated Methods", "Compute Cores Files"). Values changed vs `main` (= new clustering).
- **Detection accuracy unchanged**: 1.0 determinable, no confusion (archetypes don't touch language detection).
- **Golden masters reblessed** (both legs): all **12,727 diffs are archetype-driven** — categorized, zero structural/language drift.
- **Full engine suite: 9681 passed**; archetype parity **14/14** (repo `feature_contract_sha` updated deliberately per the intentional-change gate).

## Notes
- The `analysis_lens.py` diff is large because the two model dicts are machine-generated and were reformatted on regeneration — the semantic change is just those two constants.
- Corpus must be re-scanned with this engine for the new archetype values to appear in output DBs (old DBs don't self-update).
- Reproducible trainer artifacts (`.txt`/`.csv`/`.json`) land in a companion `gitgalaxy-population-analyses` PR.


squid-protocol:
## Process notes — this retrain took far longer than it should have

Recording the friction so the next archetype retrain is a fraction of the effort. Ranked by payoff.

### 1. The corpus DB was on the HDD → the rollup was I/O-bound (biggest single cost)
`master_db_updater.py --update function_clusters/rollup_function_clusters` does millions of SQLite `UPDATE`s on the ~5.5 GB corpus DB. On the spinning disk this was a seek storm — identical 100k-row chunks took **4 s to 124 s** (30× swing = disk, not CPU), and the function rollup alone took **~43 min**. Moving the DB to NVMe cut the *entire* files→apply→rollup→freeze tail to **~3.5 min**.
- **Fix:** stage `master_db_v<v>.db` on NVMe (or any SSD) before the rollup, not after. Ideally the trainer/updater should `PRAGMA`-tune for bulk writes (one transaction, `journal_mode=OFF`/`MEMORY` during the batch update) and/or warn when the DB is on rotational storage.

### 2. There is no single "retrain" entry point — it's ~10 manual, order-sensitive steps
The real order is: merge → train functions → **label functions** → `master_db_updater function_clusters` → `rollup_function_clusters` → re-fit files → `apply_file_clusters` → `rollup_file_clusters` → freeze → (per-language) → bake. The skill documents a simplified version that diverges from this (e.g. it doesn't mention that `func_cluster_ratios` must exist before the file fit/freeze, which cost a wasted premature file fit). 
- **Fix:** commit a single `retrain.sh` / Makefile target that runs the whole tower with `set -e` and per-stage verification. I hand-rolled exactly this (`retrain_master.sh`) — it should live in the repo.

### 3. The trainer doesn't emit `DNA_SOURCES`, but the engine needs it → the bake is a manual merge
`DNA_SOURCES` maps each `log_density_<x>` feature to its runtime `hit_vector` key, and 31/33 (func) + 55/55 (file) of those are **non-identity remaps** (`struct_branch → branch`). The trainer reads DB columns by feature-name directly so it never needs the mapping — but pasting the emitted brain over the engine constant drops it, and the engine then reads zeros → garbage/`Unclassified`. I carried it over from the prior brain by hand.
- **Fix:** make the trainer emit `DNA_SOURCES` (it's schema-stable; the mapping lives in the feature-governance config). Then the bake is a clean paste, matching the skill's "self-describing brain" ideal that functions/files are *supposed* to follow.

### 4. `cluster_names` re-mapping across a re-fit is manual and easy to get wrong
k-means cluster IDs aren't stable across fits, so the 14 curated function names have to be re-mapped to the new centroids. I did it with an optimal (Hungarian) assignment on the aligned FEATURE_NAMES contract — which also surfaced a real taxonomy shift (a split + a merge) that needed a human call.
- **Fix:** a `--relabel-from <old_brain>` flag on the function trainer that does the nearest-centroid remap and prints the mapping + low-confidence/ collision warnings for review.

### 5. Smaller papercuts
- **Inconsistent flags:** `cluster_functions.py --clusters` (plural) vs `cluster_files.py --cluster` (singular). Standardize.
- **The bake itself is a hand-edit of a 5000-line file.** I wrote a `bake.py` (splice the two model dicts by brace-matching, merge DNA_SOURCES, copy the frozen JSONs, print the new `feature_contract_sha` to pin). Worth committing next to the skill.
- **`pprint` output isn't `ruff format`-clean** — the regenerated model dicts failed ruff on first push. The bake script should `ruff format` the file as its last step.
- **The golden re-bless + `feature_contract_sha` bump are always required after a retrain** (12,727 archetype-driven golden diffs here) — bundle them into the retrain flow rather than discovering them at PR CI time.

### Net
The clustering itself was fast and clean (freeze self-validated at 100%). Nearly all the wall-clock went to (a) the HDD I/O and (b) reconstructing an end-to-end runbook the tooling doesn't encode. Items 1–3 would turn a multi-hour, error-prone session into roughly a one-command, ~15-minute run.

